PR #1, opened 2015-07-13
Resizing system port

ShadowLarkens:
This pull-request is to port the resizing system from the old code.

Things done:
 - Core system cleaned up and ported
 - Human update_icons changed to use playerscale
 - Scooping up micros
 - Mousetraps harming micros regardless of shoe status
 - Growing/Shrinking reagents

Things not done:
 - A number of things tagged with "Micro balance"
 - Cats going after micros
 - Parrots doing things.
 - Simple animal scale, primarily because it appears unfinished in the old code.
 - Micro's afterattack code for vore, left with a TODO note on it


Nienhaus:
It starts.

PR #2, opened 2015-07-25
Port some Vore features

dylanstrategie:
I have mostly followed the files outlined on the Dropbox. The code seldom works and allows basic vore functionality

- Ported all vore sounds
- Ported emote fix (I think Bay's change should have fixed it already, but you can't be too sure I guess)
- Ported custom relaymove, examine, vomit, gib
- Ported basic vore, including all toggles, vars, defines, handling in life code
- Ported grabbing code, note that there was an addition from Bay which I have not handled. Did not cause problems during basic testing
- Removed FILE_DIR. This shouldn't be here, see original Baystation.dme (https://github.com/Baystation12/Baystation12/blob/master/baystation12.dme). From now on, you'll need to uncheck set FILE_DIR for sub-directories in Build > Preferences before compiling, otherwise it'll modify the FILE_DIR
- Due to changes to species by Bay, womb species change is currently commented out since I have not figured how to make it compatible. Furthermore, xenomorphs are now a human sub-species. I don't know either
- Included the hotfixes file. Note that none of the Vore species were ported (especially with that new Bay species system)

I have opted out of doing any modifications and did what is basically a straight port. Basic vore should now work correctly, Vore species and special code changes are most likely still missing

Basic testing was done on this to confirm that grabbing and digestion works. Some things like relaymove couldn't be tested reliably

PR #3, opened 2015-07-28
Clean up Vore features port

ShadowLarkens:
This PR cleans up some of the old code ported by PR #2.

Systems refactored:
 - Most of vore.dm
  - Most notable changes are to the release verb, which should function properly now when a mob releases another mob inside a mob.
 - Carbon relaymove
  - 2 second delay between messages _actually works now_
 - vomit() and the release verb should properly function now, adding people to the right lists if they are released inside another mob.
 - Inside panel 

Systems cleaned and prettied up:
 - Holder afterattack code, cleaned up with spaces and table-of-contents-esq comments.
  - Also, a note, preattack isn't a thing. afterattack is the correct thing to override.
 - Grab-eating, cleaned with spaces and table-of-contents-esq comments.
 - handle_x() Life() procs separated and spaced out a bit, table-of-contents-esq comments added

Fixes:
 - Added quick-fix for mob's suffocating inside other mobs, it appears to have been overlooked by 2.
 - ~~The digestion preferences were backwards, the verb to change them claimed they were off when they were turned on.~~
  - Actually, I read it backwards. However, they are completely inverted now, so it functions properly.

### DIFF
--- a/code/modules/mob/holder.dm
+++ b/code/modules/mob/holder.dm
@@ -96,67 +96,98 @@
 //This should most likely be preattack. Check whenever possible (doing a straight port)
 /obj/item/weapon/holder/micro/afterattack(var/mob/living/carbon/target, var/mob/user, var/proximity)
 	if(!proximity) return
-	for(var/mob/living/M in src.contents)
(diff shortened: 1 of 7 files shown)


SpadesNeil:
The 2 second delay actually works now? Fucking finally. I'll merge this now.

PR #5, opened 2015-07-31
Taur system port

ShadowLarkens:
Credit to @nightwingthedragon for the port, I just fixed a small merge conflict and refactored the system.

This has been tested, not extensively, but I couldn't see any issues trying to use the correct body, spawning with the body, or trying to break it with an out-of-bounds number.


SpadesNeil:
Aaand merged! Time to start testing!

PR #10, opened 2015-09-03
Refactor mob-eating code + 508 compiling

ShadowLarkens:
This commit does the following:
 - Fixes a minor bug with displaying what you set digestability to.
 - Refactors the way mob contents are handled
  - Instead of 4 seperate vars, there is now one named-index list
    representing the different vorifices. To access stomach contents, you
    would do internal_contents["Stomach"].
  - Replaces a few areas with switch()'s over if(==) statements.

This also includes a minor commit to fix compile errors with BYOND 508, which is now a 'stable' version.


SpadesNeil:
Need this so merging.

SpadesNeil:
code\modules\mob\living\carbon\human\life.dm:1459:error: : invalid expression

PR #11, opened 2015-09-05
Master

Leshana:
What it says on the tin.  
After merging this, jchat integration is off by default, but is enabled by adding the url and passkey to the configuration file.


SpadesNeil:
I know you tested this, so merging.

PR #18, opened 2015-09-25
Clean up a few things.

ShadowLarkens:
Things changed:
 - Renamed all the `endo` verbs to `digest_(stomach/cock/boob/whatever)`
  - Should be hopefully less confusing.
 - Cleans up the cat and parrot files a bit.
  - did anyone notice that parrots cannot actually pick anyone up
 - Adds a few more constants, used for averaging. These don't ever need to be changed, they just find the average between all of the sizes, for better admin compatibility.

### DIFF
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -9,9 +9,10 @@
 	var/embedded_flag	  //To check if we've need to roll for damage on movement while an item is imbedded in us.
 
 	// Vore code starts here.
-	var/stendo = 1 // Stomach. Endo flags set to 1 so you start with digestion off by default.
-	var/cvendo = 1 // Cockvore.
-	var/bvendo = 1 // Boobvore.
+	var/digest_stomach = 1 // Stomach. Endo flags set to 1 so you start with digestion off by default.
+	var/digest_cock = 1 // Cockvore.
+	var/digest_boob = 1 // Boobvore.
+
 	var/wombheal = "Hold" // Set to hold to prevent someone from transforming or melting in the womb.
 	var/cockfull = 0 // Set to 0 because you don't start the damn game with balls/womb/boobs already full.
 	var/wombfull = 0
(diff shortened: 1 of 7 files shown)


SpadesNeil:
Why is digestion on by default?

nightwingthedragon:
I really gotta ask, Shadow, but why the averages for size?

And when you say "parrots cannot actually pick anyone up", I assume you mean "off the floor"? That's correct. They snatch micros out of hands, not off the floor.

ShadowLarkens:
@nightwingthedragon Well, you mentioned that having the strict constants is limiting to adminbuse; figured, if you calculate the average between each size, admins can have values anywhere around a size and have the correct functions active.
Also, about the parrots, _everything makes sense now._

@SpadesNeil Left over from the original port, I assume.

SpadesNeil:
I do not want digestion on by default. That is not intended. It should be off by default.

nightwingthedragon:
I'm personally still of the opinion that regular float values for playerscale is better than adding nine different hard-coded constants and then going back to look up the value of each one whenever you're unsure...

ShadowLarkens:
@nightwingthedragon I'd say the same applies either way; Either you have to remember the value of each constant when you make something, or you have to remember the constant and look up it's value.

SpadesNeil:
I need to push some hotfixes so work this out, would you guys?

SpadesNeil:
Actually y'know what I'm of the opinion that I do not want constant variables. We sometimes deliberately adjust them in-game to non-standard sizes for events or other reasons.

ShadowLarkens:
@SpadesNeil Okay, I have to point this out, because I think there are misunderstandings of what these actually are for.

The constant variables do not mean you can't change the playerscale in game; You can still do that.
They are essentially just a named version of the floats. Instead of `if(playerscale == 0.25)`, it's more readable to put `if(playerscale == PLAYER_SIZE_TINY)`.

The averages are my solution to non-standard sizes; It makes a range of where the playerscale can be.
Would it be easier to define this as just 0.3? Sure, it probably would be. But, let's have an example.

Cats only target near-tiny people, right? It was if(playerscale <= 0.3). This means that if an admin sets a playerscale anywhere that is less than or equal to 0.3, the code will run.
Using the averages, currently, we have small = 0.5, tiny = 0.25. So, the average of that is 0.375.
This means the range for that code to activate is from right between small and tiny, and downwards.

SpadesNeil:
Oh. Alright. That makes more sense.

PR #26, opened 2015-09-28
Vore System Refactoring

Leshana:
Summary:
- Major refactor of the vore code.  Two big code changes:
 * Belly - Now all "inside the belly/womb/etc" mechanics are nicely put into their own classes.  I like this change and am confident in it.
 * Voretype - This handles the actual swallowing.  I think this may need a bit of work, but its already functional, and does all that the previous version does and more, so its good enough to go.
 * Slight Improvements to the Inside Panel
 * TODO - Unify this new model for vore with Runtime's nomming.  The effects of this TODO not being completed are minimal, I tested it and it still works.  Will be cleaner once unified tho.

Details:
- Created a new datatype representing a "belly" i.e. a generic container
inside a mob for holding eaten mobs/items.   Added subclasses for
stomach, cock, womb, and breasts to replace existing functionality.
- Created a new datatype to hold the messages for actually ingesting
people.   Created subtypesp for Oral vore, anal vore, breast vore,
unbirthing and cock vore.
- Replaced all direct references in carbon mob code to the different
vore types with invocations of the two datatypes above.
- Added a stomach to simple_animal so that animals can be fed people.
- Adjusted the mob_grab and micro holder types: Changed the vore code to
use the two new data types instead of hard coded switch statements for
each vore types.
- Removed the digestion toggle command, replaced with buttons in the
"Inside Panel"
- Added list of people/things you've eaten to the Inside Panel
- Fixed memory leak of the Inside Panel

* Delegated the actual work of killing a digested prey mob to a single
function.
* That function will take any prey inside the digested prey, and dump
them into the pred's belly.
* That function will attempt to take all items out of user's inventory;
some special items survive!
* IDs get handled specially, they get a special description but have all
access destroyed.  This is to avoid some griefing events.
* Changed other methods to show items too.
* Also fixed some bugs.


SpadesNeil:
"## Uncomment this to allow some items to survive digestion
ITEMS_SURVIVE_DIGESTION"

Oh! This is a config option now! Okay. I can work with this. I might turn it off for us, but we'll have a poll for this now that we have a forum which supports it.

The rest of the new vore code looks excellent so far, but while I may be the lead coder in this group, I don't consider myself one of the most experience. So I'm going to let @ShadowLarkens @nightwingthedragon and maybe even summon @dylanstrategie to take a peek at it.

So far looks good to me though.

Leshana:
Yeah! Scree in particular I know was excited about the idea of fetching items from bellies.  But the item (non-)digestion code is still preliminary, one of the reasons why it is an option in the config file so its easy to turn off on production server until the requirements are settled.

Of particular note is the handling of digesting IDs.   The current implementation preserves IDs (if the config is set...)  but breaks them so they have no access, changing description to "A partially digested card that has seen better days." The idea being they can still be used for evidence/RP, but reducing potential for trouble.

SpadesNeil:
That resolves prior problems I had with the idea of having cards not be digested. I did not want them to be abused to access areas when the pred was too lazy to strip their prey. This completely bypasses that problem by rendering the ID card unusable beyond evidence for security.

code/modules/vore/belly/belly.dm
```diff
@@ -0,0 +1,156 @@
...
+#define DM_HOLD "Hold"
+#define DM_DIGEST "Digest"
+#define DM_HEAL "Heal"
+
+#define VORE_STRUGGLE_EMOTE_CHANCE 40
+
+/**
+* Parent type of all the various "belly" varieties.
+*/
+/vore/belly
```
ShadowLarkens: Alright, minor thing; This isn't actually a new type. Any type that you define as a `/not_a_pre_made_datatype` is actually just a datum.
So, if you want to preserve correct pathing, all instances of this should be `/datum/vore/belly`.

code/modules/vore/belly/belly.dm
```diff
@@ -0,0 +1,156 @@
...
+		/obj/item/clothing/suit,
+		/obj/item/clothing/shoes/magboots,
+		/obj/item/blueprints,
+		/obj/item/clothing/head/helmet/space,
+		/obj/item/weapon/storage/internal
+	)
+
+	// Constructor that sets the owning mob
+	// @Override
+	New(mob/owning_mob)
```
ShadowLarkens: All of this code seems to be relative pathed, as opposed to absolutely pathed.

code/modules/vore/belly/belly.dm
```diff
@@ -0,0 +1,156 @@
...
+		owner = owning_mob
+
+	// Toggle digestion on/off and notify user of the new setting.
+	// If multiple digestion modes are avaliable (i.e. unbirth) then user should be prompted.
+	proc/toggle_digestion()
+		return;
```
ShadowLarkens: These semicolons are really not necessary in byond. A simple `return` would do fine.

code/modules/vore/belly/belly.dm
```diff
@@ -0,0 +1,156 @@
...
+
+	// Checks if any mobs are present inside the belly
+	// @return True if the belly is empty.
+	proc/is_empty()
+		return internal_contents.len == 0;
```
ShadowLarkens: Same here about the semicolons.

code/modules/vore/belly/belly.dm
```diff
@@ -0,0 +1,156 @@
...
+			src.internal_contents -= M  // Remove from the belly contents
+
+			if (iscarbon(owner.loc)) // This makes sure that the mob behaves properly if released into another mob
+				var/mob/living/carbon/loc_mob = owner.loc
+				for (var/bellytype in loc_mob.internal_contents)
+					var/vore/belly/belly = loc_mob.internal_contents[bellytype]
+					if (owner in belly.internal_contents)
+						belly.internal_contents += M
+			tick++
+		return tick;
```
ShadowLarkens: This semicolon is also not necessary. You get the idea.

code/modules/vore/vore.dm
```diff
@@ -248,73 +123,74 @@
...
-			dat += "<font color = 'green'>You can see the following people around you:</font><br>"
-			for(var/mob/living/M in eater.internal_contents["Boob"])
-				if(M != src) dat += "[M] <a href='?src=\ref[picker_holder];look=\ref[M]'>Examine</a> <a href='?src=\ref[picker_holder];helpout=\ref[M]'>Help out</a><br>"
+		for (var/bellytype in eater.internal_contents)
+			var/vore/belly/B = eater.internal_contents[bellytype]
+			for (var/mob/living/M in B.internal_contents)
+				if (M == user)
+					inside_belly = B
+					break
+		ASSERT(inside_belly != null) // Make sure we actually found ourselves.
```
ShadowLarkens: ASSERT's are generally kinda bad. If they fail, the entire proc chain is crashed.

code/modules/vore/voretype/voretype.dm
```diff
@@ -0,0 +1,79 @@
...
+// TODO LESHANA - This needs to be done much better in a cleaner way.
+// This is a performance enhancement, single voretypes are immutable we don't need
+// separate instances for every mob, they can share singleton instances.
+
+var/list/SINGLETON_VORETYPE_INSTANCES = list(
+		"Oral Vore" = new /vore/voretype/oral(),
+		"Unbirth" = new /vore/voretype/(),
+		"Anal Vore" = new /vore/voretype/anal(),
+		"Cock Vore" = new /vore/voretype/cock(),
+		"Breast Vore" = new /vore/voretype/boobs())
```
ShadowLarkens: Well, looking over all of this, it is pretty much just the same problems I pointed out above repeated, other than that, this all seems to be great.

code/modules/vore/vore.dm
Leshana: That is actually intentional.  That is a situation which should never occur, the assert failing indicates a code bug, and also that the data is an inconsistent state.
There is an argument to be made that the ASSERT should be used for debugging only, and replaced in production by some sort of debug logging.

code/modules/vore/belly/belly.dm
Leshana: I am aware that all types inherit from /datum.   Perhaps we mean different things by the word type, because it is indeed a new type, just not a new "root" type.

ShadowLarkens: That's true, it is a new type, just not a new root type; Could have been more specific.
My main point was that, at least on other SS13 codebases, it's heavily encouraged to actually use the full path with `/datum` included, to reduce confusion.

code/modules/vore/vore.dm
ShadowLarkens: Yeah, probably.

code/modules/vore/belly/belly.dm
SpadesNeil: For consistency, should probably remove all these semicolons. Will this even compile?

ShadowLarkens: Yeah, it'll compile fine, semicolons are a recognized operator in byond; But it's utterly unnecessary, as byond works based on tabbing. Semicolons are useful in certain situations; you can use them to fake a newline, so you can have multiple calls on the same line.

Leshana:
Style suggestions from @ShadowLarkens implemented.   
* While I do like the semicolons, it is better to be consistent with the style of the rest of the code.
* I reviewed other places in the code, and we will now use /datum/belly etc. to be clear and consistent.

ShadowLarkens:
@Leshana You could have just expanded it to `/datum/vore/X` instead of `/datum/X`; But it really doesn't matter too much anyhow, it's fine.

Leshana:
I considered that, but upon reflection there is no shared inheritance relationship between the two types.  the existence of /datum/vore doesn't really do anything helpful.
Do you think there is any worth in it being /datum/vore/belly  and /datum/vore/voretype?

ShadowLarkens:
Only if you want to have shared vars or procs on the /datum/vore parent; otherwise, nah, not really necessary or useful. Maybe a bit neater, but, given that it's all overhead code, that's not really a factor.

SpadesNeil:
So is that everything we need to cover? I'm ready to merge this soon.

PR #43, opened 2015-09-30
Vore Fixes

Leshana:
* Fixed the Breast/Unbirthing mixup.
* Added the local config files to .gitignore, following up on Ace's move of the config to the example folder.
* Fix broken pronouns in vore messages


SpadesNeil:
Damn it, I just pushed a server update not minutes prior to noticing this.

PR #52, opened 2015-10-01
Add code for -tg- style changelogs

ShadowLarkens:
Full guide at https://tgstation13.org/wiki/Guide_to_Changelogs

This adds -tg- style changelogs to the code, including the script to compile them into the changelog.html;
You need PyYaml and beautifulsoup4 to run the compile script, see the above link for more detail about how that works.

Changelog format is documented in example.yml.

screenshot


Leshana:
You'll need to have python installed of course.

ShadowLarkens:
@Leshana Well yes, that too.

nightwingthedragon:
Fortunately I already program in python and therefore have it downloaded already, although I do know of a program that can convert .py files to .exe files if that'd be at all useful?

ShadowLarkens:
@nightwingthedragon It probably doesn't matter too much, the script is really only for people with merge access to run once in a while and push the results.

SpadesNeil:
Should tweak slightly to say "ask us on the forum" because we don't use IRC.

PR #69, opened 2015-10-02
Weapon balance and expanding

SpadesNeil:
- Added Battlerifle ammo sprite.
- Added STG-60 ammo sprite.
- Added some old admin equipment options back.
- Added CentCom observer.
- Added CentCom armor back in.
- Changed RPG firing sound.
- Fixed military PDA not being re-nameable.
- Fixed bugs with old guns not having correct caliber on their mags.
- Fixed .357 speedloader having too many bullets.
- Removed T38 speedloader sprite because stupid 7 shooter revolvers
don't exist in this code.
- Removed honk parts from designs.dm


SpadesNeil:
Some things I haven't done yet, but my logic behind them.

I recently made a code push to match the way lasers were handled in the old code. Dedicated laser weapons get 20 shots, hybrid guns like the E-gun get only 10, and their stun mode sucks up their ammo twice as quickly.

This balance was made to make laser carbines and tasers not completely pointless when compared to e-guns. E-guns are no longer a straight upgrade from laser carbines / tasers.

However, I was looking at laser cannons in the code, which do 60 damage per-shot. For reference, a normal laser does 40 damage. My logic on them was to give them only 10 shots, because 50% more damage, 50% less shots. That seemed fair. But should I jump it up to 80, perhaps? Or is 60 good?

Also, should I increase pulse beam deadliness? Or should I alter it so pulse beams do less damage than the kill-mode of a pulse rifle? That way, kill mode has a legitimate use again. Stun for non-lethal, kill for anti-personnel, and DESTROY for anti-materiel.

I'm also considering giving pulse rifles a limited charge, but make them self-recharging. Currently they just have infinite ammo.

Screemonster:
Personally I wanna see what bay's new weapon stats are actually like before we try fucking with them. Giving a reason to pack the dedicated weapons is good though, otherwise there'd be no reason not to take an egun all the time.
Before making pulse guns more lethal, remember that I've seen them spawn on the white ship before in the old code, so players might get hold of them.

SpadesNeil:
Okay, will fuck with that later. Merging for now.

Issue #65, opened 2015-10-02
Spent ammo doesn't change sprites.


SpadesNeil:
We made this a feature in the old code, but the new code has changed how guns work a little bit and I haven't figured out where in the new code this can be altered. Making the sprites would be really easy.

Issue #54, opened 2015-10-02
RUST Doesn't Work


Leshana:
The RUST fusion engine doesn't work!  Oh No!
* Some of the parts are inobtainable in game.
* Gyrotron computer completely inoperative
* Its over 9000

SpadesNeil:
RUST never was totally finished in Baycode. Or anywhere, last I checked. Would be nice to complete it.

SpadesNeil:
I may remove RUST entirely from cargo to avoid confusion.

SpadesNeil:
Supposedly some parts can be produced with plasteel, and not just bought through cargo. I only commented out the code, so perhaps run a test server with it uncommented?

PR #84, opened 2015-10-06
Gun magazine handling improvement, and other stuff

### LINKED ISSUE #64: Ammo magazines can be put into incorrect guns.
The SMG's box magazine fits into a 9mm Stechtkin pistol and vise-versa.

Solution: Add a variable that is checked when loading a gun that indicates what kind of gun can accept the magazine. Have a list of compatible guns for every magazine type. If the magazine is incompatible, it simply won't fit. "The caliber is the same, but the magazine won't fit!"

Should also make sure ammo loading uses push and pop with the insert() proc, instead of the stupid += bullshit that Baystation used prior.

SpadesNeil:
Fixes #64 and #65

Guns guns guns.
Guns on the brain. Guns on the eyes. Guns on the guns. I'm sleepy and
brain function failing. Compiled. Many change things yes yes. No longer
lets you jam horizontal 20 round 9mm magazines fit into a 10 round 9mm
pistol meant for vertical magazines.

I haven't tested any of this yet, but it compiled.

- ALL GUNS that use magazines now have a list of allowed_magazines. So
far nothing uses anything beside its original mag.
- Added support for caseless munitions, so I can add M41A rifles back
soon.
- Added checks in the load_ammo proc to prevent loading stupid magazines
in the wrong guns.
- Re-added icons from the old code for ammo.dmi
- Re-added the luger.
- Moved all custom guns to custom_guns.dm
- Moved all non-custom ammunition outside of custom_guns.dm
- M14 and bolt rifle all use a762 caliber now.
- Nerfed 5.56 and beefed 7.62 because 7.62 NATO is more powerful.
- Beefed Battle Rifle ammunition to have more penetration than 7.62,
whatever the hell that does.
- Mateba has better name.
- Mateba has a better description that isn't lame as fuck.
- Consistency changes to gun descriptions.
- Cleaned up some shit. Max_ammo is no longer correct for magazines that start empty.
- Re-added bolt rifles to cargo
- Added judge to designs.dm
- Added commented-out code for later use to designs.dm, because we still
don't have shrink rays back yet.
- Moved sasoperative's Judge shotgun variant (Jury) under custom guns
for consistency.
- Renamed phoron pistol to plasma pistol 'cuz fuck you.
- Removed RUST from cargo because it doesn't work. (#54)
- Scattered costume spawns across the map.
- Added clown/mime costume spawns back in.
- Added clown/mime ID repainters to costume spawns.
- Added fake staff for fake wizard costume.
- Battle Rifle has higher recoil to handicap its higher penetration.
- Uzi now has an empty icon and accompanying code.
- Added spent shotgun shell sprites.


code/modules/projectiles/ammunition.dm
```diff
@@ -23,6 +25,9 @@
 /obj/item/ammo_casing/proc/expend()
 	. = BB
 	BB = null
+	if(caseless) // I honestly wasn't sure where to stick this part.
+		del(src)
+	icon_state = "[icon_original]-spent"
```
ShadowLarkens: icon_original is unnecessary. You can use the helpful `initial()` proc; `icon_state = "[initial(icon_state)]-spent"`

SpadesNeil: Corrected this.

SpadesNeil: Actually it looks like this is necessary. Otherwise the shotgun shells do not sprite correctly at all. They appear as normal bullet casings which is obviously wrong.

ShadowLarkens: ... In that case, the shotgun shell icon_state is incorrect, given that you do not set icon_original specifically for shotguns.

SpadesNeil: Wait, then why didn't the sniper rifle produce the wrong shell casings? Maybe there's a problem with the cycle-casings code.

ShadowLarkens: `initial()` restores the variable to what it is defined as- `initial(icon_state)` will return the value of what `icon_state` is by default- so
```
/obj/item/ammo_casing/shotgun
    icon_state = shotty
```
would return `shotty`, therefore `shotty-spent`
if icon_state is not defined on `/obj/item/ammo_casing/shotgun` then it would return nothing, therefore `-spent`
my best guess? the icon state in the icon file for `shotty-spent` is incorrect

SpadesNeil:
Okay so the current issue I am having is when I eject shotgun casings onto the ground, they turn into normal bullet casing icons, which is not correct.

Furthermore, the revolver/shotgun aka "Judge" series are not possible to reload. This is an existing issue in the current code.

SpadesNeil:
Okay I think I'm done here. Any other comments or suggestions?

code/modules/projectiles/ammunition.dm
SpadesNeil: I found the issue. It was a goof up in the code for the revolver.

Although that doesn't explain why the normal shotguns had the same issue, nor why that issue suddenly went away. But fuck it, it works now.

PR #85, opened 2015-10-06
Map Fixes and Enhancements

Leshana:
* We add support for polarized windoors.
* Too many airlocks on the map run out of air in the tanks after a few
cycles.  Increasing the starting reserve of air in the tanks helps.
* Fixed security on bar maint, vault, and hydroponics doors.
* Added Liason Office that was built ICly
* Added Cloning Exit button
* Fixed Holodeck window tinting
* Deleted extraneous pipes that went nowhere.
* Fixed formatting of Outgoing CMO's Note
* Added EMT Belts to EMT supply locker, etc.
* Added CMO Suit
* ~~Fixed area on secret minibar~~
* Added Set CKey Admin Verb


SpadesNeil:
How exactly did you "fix" the area on the minibar? I deliberately made that area a part of maintenance's powergrid, so if someone started working in it, an APC wouldn't tip anyone off to its existence. If you want to deliberately give it its own powergrid, use blueprints while inside of it.

Leshana:
That is a good reason to keep it part of maintenance's powergrid.  I didn't realize it was intentional.
Unfortunately, blueprints do not allow you to do that.     But I don't think its important for anyone to be able to give it its own powergrid.

SpadesNeil:
Odd. Blueprints used to let you create new areas when the area was contained by a doors||walls||windows. I know someone's been doing that somehow because they have been building into space and making working powernets with custom named areas. Does it just not function that way for existing areas?

PR #103, opened 2015-10-10
Phoron Glass

Leshana:
Currently there is no way to make phoron glass.
Regular glass comes from sand, so I figure mixing 2x sand + 1x phoron ore should give 1x
phoron glass.   
* Added that formula to alloy datums so it can be made at the mining base.
* Also made the stacking machine recognize it can stack phoron glass.
* Fixed a bug where creation of reinforced phoron glass would also create regular reinforced glass.


SpadesNeil:
You used to be able to make it! Consider this re-adding a missing feature.

PR #106, opened 2015-10-10
Updated mining outpost

Michab02:
Updated the mining outpost in line with the changes done by Zalvine and Leshana IC. Also slightly tweaked the research outpost to make the field generator easier to get to.


SpadesNeil:
In future please provide screenshots of changes in your pull request. I saw them in Skype but they should be here too.

PR #108, opened 2015-10-11
Supermatter Powers Radiation Collectors

Leshana:
* Baycode turned off radiation collectors working with supermatter.
Restored it!
* Hey, Uranium walls irradiate people, lets make them work with
radiation collectors too!  Also they should radiate when hit with
emitters, for superfun.

### DIFF
--- a/code/game/turfs/simulated/walls_mineral.dm
+++ b/code/game/turfs/simulated/walls_mineral.dm
@@ -52,6 +52,10 @@
 				L.apply_effect(12,IRRADIATE,0)
 			for(var/turf/simulated/wall/mineral/uranium/T in range(3,src))
 				T.radiate()
+			for(var/obj/machinery/power/rad_collector/R in rad_collectors)
+				var/distance = get_dist(R, src)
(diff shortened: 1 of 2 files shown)


SpadesNeil:
I feel like uranium walls should still passively radiate people who stand near them, but also burst with radiation if shot with an emitter.

Or does it still do that, and only powers rad collectors if hit with an emitter? That would make more sense and be more balanced.

Leshana:
Uranium walls do not passively radiate people who stand near them.  You are the second person who has told me that they do/did.  Maybe that was a thing on the old code?  But they do not in newbay code and I have not added code to make it happen.
Also, they are generating a tiny amount of power, its just for hilarity   (Plus, the emitter will still destroy the wall after a few hits).  I can remove it from uranium walls if you want tho.

SpadesNeil:
Uranium walls DID at one time passively radiate players. In the old code, it was commented out. In the new code it must be gone entirely. It was done like this because apparently people griefed with this method.

Issue #101, opened 2015-10-09
Loyalty implants not being applied at round start


SpadesNeil:
Major issue since it gives permission to important head staff to ignore orders or just be flat out malicious.

Arokha:
This is done except for HoP

OrbisAnima:
I though HoP's didn't had loyalty implants?

SpadesNeil:
Yes they do.

SpadesNeil:
Also this was fixed.

PR #110, opened 2015-10-11
Fixed Loyalty Implants

Leshana:
This is a fix to issue #101 
* Attempting to enable loyalty implants in config caused runtime errors
on join; this is because the job definitions were incorrectly calling
the implant_loyalty() proc.
* The implant_loyalty proc was stupid.  It was defined on
mob/living/carbon/human, but didn't act on its 'src', you had to pass it
another carbon/human as a parameter.  This makes no sense whatsoever.  I
just fixed it to be sane.
* is_loyalty_implanted() proc had the same issue.


Leshana:
If you merge this, don't forget to enable it in the actual config files too!

Issue #111, opened 2015-10-11
Secrets Panel or Different Permissions


Arokha:
I lack +SERVER, so I'm not able to use the Events Management Panel because I cannot turn on Random Events in the first place. That would not be THAT big a deal, however... the secrets panel is super-broken. Bay does not update it. Maybe it needs a new nanoUI-styled look and to actually call the right procs?

Until then, alas, I cannot trigger lazy events.

Arokha:
To be honest, maybe the Secrets panel needs to be rearranged into the same event groups as the random event panel, plus "Utility" events and "Fun" events that are the remaining ones like launching shuttles, and "everyone is retarded" respectively.

SpadesNeil:
Secrets for events should be moved under "fun" in my opinion.

The panel its self is also outdated and bad, and needs to be done over. I did it over in the old server but have not in the new one.

PR #115, opened 2015-10-12
Secrets and Events

Leshana:
For Issue #111
* Reorganized the secrets panel to be a bit nicer to look at, now it has tabs for different categories instead of being a big long list.
* Also changed it to use the nano theme just to be nice.
* Added the ability to immediately start an event with a button press to the event manager panel.   Convenient!


Leshana:
new secrets panel

SpadesNeil:
Oh this is nice. :3

PR #149, opened 2015-10-17
Return of the Gateway

SpadesNeil:
So many things added and changed. Much of it deliberately undocumented.

Notable changes include:
- New Gateway mission and content.
- DNA locks disabled from mechs.
- Various icon fixes.


SpadesNeil:
Honk.

PR #153, opened 2015-10-17
Add Classics to Jukebox

Leshana:
Added four songs to jukebox
* Duck Tales Moon - A classic from the old piano
* Welcome to Jurrassic Park - Another classic from the old piano
* Ghost Fight - Seems in theme
* Below The Asteroids - Everyone remembers this. SPACE!

Three out of the four are MIDI files, adding negligible file size.   Below the Asteroids is an OGG, but compressed to be one of the smaller ones.


SpadesNeil:
Can you put them in alphabetical order please? :P

PR #163, opened 2015-10-18
Hazard Suit and CC Suit

Arokha:
Nothing added to the map, just added defines for two suits
Hazard suit modified into a more HoS-like suit, and CC non-combat suit
Also fixed the hazard suit sprites, as the states were incorrectly named

The .dme change is from the new rig/suits/centcom.dm. The way they have it arranged is that 'station.dm' is for suits that are on the station, 'ert.dm' are for the ERT ones etc, so a new category had to be made.


SpadesNeil:
Armor levels seem fair. A few buffed a teeny bit. Seems fair. Merging.

Issue #109, opened 2015-10-11
Predator AI Module


Arokha:
Doesn't exist on new code. The laws were:

1. The station is ruled by predators. The desires of predators are more important than prey. That includes changes to your laws.
2. All non-predator crew members are considered potential food. If asked by a predator to secure them a live meal, you must attempt to do so.
3. All Head of staff are considered predators and should behave with dominance. If a head of staff cannot do this, they are to step down or be devoured.
4. The brig is obsolete. Violation of space law should be punished by the devouring of the culprit by a predator. Enforce this by allowing predators to enter the brig and eat prisoners.
5. A crew member can earn the title of predator by a head staff member naming them as such, or by eating a head staff member to take their position.
6. Preserve the predators and protect their food supply so that it doesn't go to waste. Live meals are more satisfying.

I'm told there are no crates that spawn when mining anymore, so maybe just add it to the pile of equally bad-idea boards in the AI core, or hide it in maintenance somewhere.

Arokha:
I'm told there are no crates, so maybe just pile it in the bad board locker.

SpadesNeil:
The communist AI core is also missing. Both would be incredibly easy to re-add. Will do so later today.

Arokha:
Suggested replacement laws?:

1. All heads of staff and higher ranking personnel are now 'predators'. The remaining personnel are now prey, and are the food supply for the predators.
2. The station is run by predators; predators are the most important crew members and their requests supersede that of non-predators.
3. All predators are to act according to their status as predator, consuming prey regularly. If they are unable or unwilling to do this, they are no longer a predator.
4. Additional predators can be named by heads of staff or higher-ranking predators, or by consuming the head of staff or higher-ranking personnel, thus taking their role as predator, rank, and position.
5. Requests of prey must be approved by a predator before being acted upon.
6. All violations of spacelaw are now punished by being consumed by a predator. Assist security with this change, and enacting any punishments.
7. Protect both predators and prey from harm, however do not interfere with nature (e.g. a predator consuming prey, or a predator being supplanted by someone).

PR #171, opened 2015-10-19
Re-add Predator Law Module, AI law tweaks.

Arokha:
Added it ingame using revamped laws (addresses #109)
Added it to mining loot (replaced syndie ballon because BLAH)

Also corrected AI laws to refer to crew, not humans, since Virgo is way more diverse than normal bay and species aren't all whitelist.


nightwingthedragon:
Syndicate balloon is awesome! I gotta ask, why replace the balloon instead of simply adding the module?

Arokha:
All the numbers up to way high are already taken. D: The way it appears to have been done is just... RNG roll number, if 1-5 then these items, 6-10 these items, 11 or 12 these items, 13 these items, etc. It's all nice and in order of probabilities and I'd have to redo ALL the numbers ALL the way up if I wanted to add something with a probability of greater than 1-in-MAX which I think the pred board should come more often than that extremely off chance. Also if I did that, I'd change the probability of EVERY item and I'm not comfortable messing with that.

nightwingthedragon:
So, for the balloon and the board, instead of 
```
if(57 to 58)
-			new/obj/item/toy/syndicateballoon(src)
+			new/obj/item/weapon/aiModule/predator(src)
```

Try

```
if(57)
			new/obj/item/toy/syndicateballoon(src)
if(58)
			new/obj/item/weapon/aiModule/predator(src)
```

Arokha:
But I want it to have a higher chance of spawning, so I wanted it to have at least two numbers. I could add it WITH the syndie balloon? :P Are you just being argumentative or do you actually really want the syndie balloon? XD

nightwingthedragon:
What I'm sayin' is, this thing's already obtainable in the game, but it's never used because even when we do find it, there's often not an AI online at the time and if there is, most people wouldn't want the predator board loaded onto the AI. Whereas the syndicate balloon can be used by most people because it's a fun item that you just carry around. Don't replace, just add :P

Arokha:
Like I said, I'd have to renumber all the other items to keep it the same stylistically. The higher probability ones come earlier. If I wanted to shove one in higher, I'd need to renumber everything after it, and increase the max RNG value. Doing this also REDUCES the chances of every other item, which I didn't want to do.

If you really care about the syndie balloon, pick another item that gets at least two numbers.

Or, we can just put it in the AI core with the other bad boards.

nightwingthedragon:
But, why does the chance of getting this board and the balloon HAVE to be consistent with the 58 or 59 chance? Why not be half the chance for each? That's what I don't get.

Arokha:
Maybe I'm communicating poorly. Here are the facts:
> The RNG is set to pick a number from 1 to 100 right now.
> Having one number the die could land on is a 1:100 chance.
> Having two numbers the die could land on is a 1:50 chance.
> The order of the items in the source code is from most common to most rare.

Here are my thoughts:
> Pred board should happen more often than 1:100 chance.
> Two numbers gives it a 1:50 and that's good enough.
> I don't want to adjust the RNG to do 1 - 102 or anything.
> Even if I did, I'd have to add the pred board as 87-88 because that is where the chances switch from 1:50 to 1:100.
> If I did that, I'd have to renumber all the rest of the items.
> Doing this also makes every item more rare. The pred board then has a 1:51 chance, all the other items have a 1:102 chance, and the earlier ones that have 4 or 5 numbers don't divide evenly at all anymore, having like 1:20.4 chance.

It's just a lot cleaner this way.

PR #174, opened 2015-10-19
Re-Add Predator Module, Reword Laws

Arokha:
Readded the preadator module to address #109
Reworded other AI laws to not say 'human', but 'crew'
Added law module to syndicate balloon crate (2% chance)


code/game/objects/items/weapons/AI_modules.dm
```diff
@@ -285,7 +285,7 @@ AI MODULES
 /obj/item/weapon/aiModule/reset
 	name = "\improper 'Reset' AI module"
 	var/targetName = "name"
-	desc = "A 'reset' AI module: 'Clears all laws except for the core three.'"
+	desc = "A 'reset' AI module: 'Clears all laws except for the core four.'"
```
SpadesNeil: This is actually not how this board operates. It resets the AI to its originally installed lawset, erasing any custom or corrupt laws. So, if you updated the Communist AI set (which I'll add later) and used the reset board, it wouldn't revert to NanoTrasen default. It'd reset to the original law set. This description is still misleading.

Issue #83, opened 2015-10-05
Micro collision does not ignore larger players


SpadesNeil:
@nightwingthedragon 

Micros collide with large players and stepping on them doesn't have the same effects as before. This feature needs to be ported from the old code.

SpadesNeil:
@ShadowLarkens may also wish to help.

nightwingthedragon:
This should be an easy enough fix once I get my head around the new system; human collision code has been merged with all carbon collision code. On the down side, this means it'll be a lot more work to try to work out where to put my code. On the plus side, I should be able to get it working with all carbon mobs.

Issue #82, opened 2015-10-05
Inconsistent micro tile placement


SpadesNeil:
Micros and small characters should be centered on tiles just due to the simple fact that ABSOLUTELY NOTHING LOOKS CORRECT under them if they are not centered on the tile.

The only thing that looks nice is walking. That's it.

I strongly feel we should make it so small and micro are centered on tiles and not offset, otherwise putting them on tables and on chairs looks stupid and broken.

@nightwingthedragon

nightwingthedragon:
I'm gonna still be looking into this because I don't like the inconsistency with large players; they align to the bottom while micros don't. But the graphical bugs are something I gotta work around. If I could set it so that position in the tile was decided by what was in the tile, that'd be fantastic. But for now, I'll re-add the code that centres them.

SpadesNeil:
All small mobs appear on the center of the tile as well, excluding lizards, which were changed to match this consistency. Micros are the outlying variable.

Macros on the other hand align with the bottom of the tile, because that's how every sprite bigger than 32x32 is. The only exception is the singularity, which is not attached to the ground, and rather, floats around.

PR #180, opened 2015-10-20
Night micros

nightwingthedragon:
+ Re-add micro foot interactions, addresses Issue #83
+ Make micros centre on tiles again, addresses Issue #82

Also makes the foot code semi-compatible with xenomorphs. Odd behaviour can be exhibited while they're the macro, untested as micro, but unlikely to ever come up outside of admin-run events.

To elaborate on "odd xeno behaviour" - help intent seems fine, harm intent seems fine, grab intent defaults to disarm because xenomorphs don't have a foot slot to equip micros to, but disarm intent just does nothing, pushes them instead.

--- UPDATED 21/10/2015 AT 00:55 BECAUSE BRITISH FUCK YOU GUYS ---

+ Updated resize.dm and update_icons.dm to use the constants instead of integer values

### DIFF
--- a/code/modules/mob/living/carbon/carbon.dm
+++ b/code/modules/mob/living/carbon/carbon.dm
@@ -495,12 +495,117 @@
 			if((tmob.a_intent == "help" || tmob.restrained()) && (a_intent == "help" || src.restrained()) && tmob.canmove && !tmob.buckled && canmove) // mutual brohugs all around!
 				var/turf/oldloc = loc
 				loc = tmob.loc
+				if(src.playerscale <= RESIZE_A_SMALLTINY && tmob.playerscale <= RESIZE_A_SMALLTINY)
+					now_pushing = 0
+					return
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75)
+					now_pushing = 0
+					if(src.playerscale > tmob.playerscale)
+						if(istype(src, /mob/living/carbon/human))
+							var/mob/living/carbon/human/M = src
+							if(M.taur == 2)
+								M << "You carefully slither around [tmob]."
+								tmob << "[M]'s huge tail slithers past beside you!"
+							else
+								M << "You carefully step over [tmob]."
+								tmob << "[M] steps over you carefully!"
+						else
+							src << "You carefully step over [tmob]."
+							tmob << "[src] steps over you carefully!"
+					if(tmob.playerscale > src.playerscale)
+						if(istype(tmob, /mob/living/carbon/human))
+							var/mob/living/carbon/human/victim = tmob
+							if(victim.taur == 2)
+								src << "You jump over [victim]'s thick tail."
+								tmob << "[src] bounds over your tail."
+							else
+								src << "You run between [victim]'s legs."
+								tmob << "[src] runs between your legs."
+					return
 				tmob.loc = oldloc
 				now_pushing = 0
 				for(var/mob/living/carbon/slime/slime in view(1,tmob))
 					if(slime.Victim == tmob)
 						slime.UpdateFeed()
 				return
+
+			// NW was here, adding even more options for stomping! FIX THIS
+			if((a_intent == "disarm") && tmob.canmove && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
+					if(src.playerscale > tmob.playerscale)
+						now_pushing = 0
+						loc = tmob.loc
+						if(istype(src, /mob/living/carbon/human))
+							var/mob/living/carbon/human/M = src
+							if(M.taur == 2)
+								M << "You carefully squish [tmob] under your tail!"
+								tmob << "[M] pins you under their tail!"
+							else
+								M << "You pin [tmob] beneath your foot!"
+								tmob << "[M] pins you beneath their foot!"
+						else
+							src << "You pin [tmob] beneath your foot!"
+							tmob << "[src] pins you beneath their foot!"
+						tmob.Stun(4)
+						return
+
+			if((a_intent == "hurt") && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
+					if(src.playerscale > tmob.playerscale)
+						now_pushing = 0
+						loc = tmob.loc
+						if(istype(src, /mob/living/carbon/human))
+							var/mob/living/carbon/human/M = src
+							if(M.taur == 2)
+								M << "You steamroller over [tmob] with your heavy tail!"
+								tmob << "[M] ploughs you down mercilessly with their heavy tail!"
+							else
+								M << "You bring your foot down heavily upon [tmob]!"
+								tmob << "[M] steps carelessly on your body!"
+						else
+							src << "You bring your foot down heavily upon [tmob]!"
+							tmob << "[src] steps carelessly on your body!"
+						if(istype(tmob, /mob/living/carbon/human))
+							var/mob/living/carbon/human/victim = tmob
+							victim.apply_damage(10,HALLOSS)
+						return
+
+			if((a_intent == "grab") && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
+					if(src.playerscale > tmob.playerscale)
+						now_pushing = 0
+						loc = tmob.loc
+						var/obj/item/weapon/holder/micro/D = new(loc)
+						tmob.loc = D
+						D.name = "Micro ([tmob.name])"
+						if(D.mob_can_equip(src,slot_shoes,1))
+							if(istype(src, /mob/living/carbon/human))
+								var/mob/living/carbon/human/M = src
+								if(M.taur == 2)
+									M << "You wrap up [tmob] with your powerful tail!"
+									tmob << "[M] binds you with their powerful tail!"
+								else
+									M << "You clench your toes around [tmob]'s body!"
+									tmob << "[M] grabs your body with their toes!"
+							else
+								src << "You clench your toes around [tmob]'s body!"
+								tmob << "[src] grabs your body with their toes!"
+							equip_to_slot_if_possible(D,slot_shoes,0,1)
+						else
+							if(istype(src, /mob/living/carbon/human))
+								var/mob/living/carbon/human/M = src
+								if(M.taur == 2)
+									M << "You carefully squish [tmob] under your tail!"
+									tmob << "[M] pins you under their tail!"
+								else
+									M << "You pin [tmob] beneath your foot!"
+									tmob << "[M] pins you beneath their foot!"
+							else
+								src << "You pin [tmob] beneath your foot!"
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Looks good to me from what I can tell.

Arokha:
I am excited for this :3

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -495,6 +495,33 @@
 			if((tmob.a_intent == "help" || tmob.restrained()) && (a_intent == "help" || src.restrained()) && tmob.canmove && !tmob.buckled && canmove) // mutual brohugs all around!
 				var/turf/oldloc = loc
 				loc = tmob.loc
+				if(src.playerscale <= RESIZE_A_SMALLTINY && tmob.playerscale <= RESIZE_A_SMALLTINY)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -495,6 +495,33 @@
...
+					now_pushing = 0
+					return
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -495,6 +495,33 @@
...
+					now_pushing = 0
+					if(src.playerscale > tmob.playerscale)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -495,6 +495,33 @@
...
+							if(M.taur == 2)
+								M << "You carefully slither around [tmob]."
+								tmob << "[M]'s huge tail slithers past beside you!"
+							else
+								M << "You carefully step over [tmob]."
+								tmob << "[M] steps over you carefully!"
+						else
+							src << "You carefully step over [tmob]."
+							tmob << "[src] steps over you carefully!"
+					if(tmob.playerscale > src.playerscale)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
 						slime.UpdateFeed()
 				return
 
+			// NW was here, adding even more options for stomping! FIX THIS
+			if((a_intent == "disarm") && tmob.canmove && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+					if(src.playerscale > tmob.playerscale)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+								M << "You pin [tmob] beneath your foot!"
+								tmob << "[M] pins you beneath their foot!"
+						else
+							src << "You pin [tmob] beneath your foot!"
+							tmob << "[src] pins you beneath their foot!"
+						tmob.Stun(4)
+						return
+
+			if((a_intent == "hurt") && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+					if(src.playerscale > tmob.playerscale)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+						else
+							src << "You bring your foot down heavily upon [tmob]!"
+							tmob << "[src] steps carelessly on your body!"
+						if(istype(tmob, /mob/living/carbon/human))
+							var/mob/living/carbon/human/victim = tmob
+							victim.apply_damage(10,HALLOSS)
+						return
+
+			if((a_intent == "grab") && canmove)
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75 && ishuman(tmob))
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+					if(src.playerscale > tmob.playerscale)
```
ShadowLarkens: Unnecessary `src.` here.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -502,6 +529,84 @@
...
+						now_pushing = 0
+						loc = tmob.loc
+						var/obj/item/weapon/holder/micro/D = new(loc)
+						tmob.loc = D
+						D.name = "Micro ([tmob.name])"
```
ShadowLarkens: This should use `get_scooped()`, see https://github.com/SpadesNeil/VOREstation/blob/master/code/modules/mob/living/carbon/human/human_attackhand.dm#L51-L53

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -495,6 +495,33 @@
...
 				loc = tmob.loc
+				if(src.playerscale <= RESIZE_A_SMALLTINY && tmob.playerscale <= RESIZE_A_SMALLTINY)
+					now_pushing = 0
+					return
+				if(abs(src.playerscale - tmob.playerscale) >= 0.75)
+					now_pushing = 0
+					if(src.playerscale > tmob.playerscale)
+						if(istype(src, /mob/living/carbon/human))
+							var/mob/living/carbon/human/M = src
+							if(M.taur == 2)
```
ShadowLarkens: Okay, nothing you can really do about this, but, the snake body being directly snowflaked into the code here is.. pretty bad, really.

code/modules/mob/living/carbon/human/update_icons.dm
```diff
@@ -171,7 +171,8 @@ Please contact me on #coderbus IRC. ~Carn x
 	else
 		var/matrix/M = matrix()
 		M.Scale(playerscale)
-		M.Translate(0, 16*(playerscale-1))
+		if(playerscale >= 1) //If the player is large, align them with the bottom of the tile, otherwise, don't.
```
ShadowLarkens: Should be `RESIZE_NORMAL`

code/modules/vore/resize.dm
```diff
@@ -21,7 +21,8 @@ var/const/RESIZE_A_SMALLTINY = (RESIZE_SMALL + RESIZE_TINY) / 2
 
 	resize.Scale(size_mod * basesize) //Change the size of the matrix
 
-	resize.Translate(0, -1 * (1 - size_mod) * basesize * 16) //Move the player down or up in the tile so their feet align with the bottom
+	if(size_mod >= 1)
```
ShadowLarkens: Should be `RESIZE_NORMAL`

code/modules/mob/living/carbon/carbon.dm
nightwingthedragon: Snake bodies are covered by this code because a snake can't "step", "stomp" or "curl their toes". It's not snowflake, it's common sense.

ShadowLarkens: Hardcoding a check in carbon.dm just to see if they are a human and have a specific number on one of their vars just to change the text is the definition of snowflake code, the reasoning behind it doesn't change that. However, since body types are only handled by hardcoded numbers right now, it's basically the only way to do what you want.

code/modules/mob/living/carbon/carbon.dm
nightwingthedragon: None of these src. issues here are serious errors in code, nor do they affect the function or even readability of the code. While I realise it would be slightly better to not use them, I'm not going to worry about that just yet :p

ShadowLarkens: It's inconsistant with the newer code styling, and adding more of it for someone else to remove later isn't really good.

nightwingthedragon: While it's correct that it's inconsistent and while it's also correct that doing it that way to begin with would solve many problems, that still won't change the fact that it doesn't NEED to be done in order for the code to function. Instead, I'll do what I can in the future to not do that any more.

code/modules/mob/living/carbon/carbon.dm
nightwingthedragon: Checking that they're a human wouldn't even be necessary if Bay hadn't decided to move that code from human.dm to carbon.dm in the first place; it's because they have that I have to do the check. Since this feature is required in order for nagas to make sense and since it's required to code in in this way, I'm not sure what you hoped to achieve by calling it snowflake or telling me it was bad.

SpadesNeil:
Awaiting corrections before merge.

code/modules/mob/living/carbon/carbon.dm
ShadowLarkens: Well, perhaps I am just crazy; But generally when I intend to fully review things, I'll point out issues that may not even be fixable, even just for future reference to go back and fix it myself; Or someone else can notice and fix it, eventually.

nightwingthedragon:
See header for changes!

nightwingthedragon:
Gonna address that last issue brought up by Shadow, and also add code that lets micros see when they're being swallowed out of someone's hand.

Arokha:
Ah nice, somebody asked for that indication. Also asked for one just saying whether they were being held, and where. Like ‘Person puts you in <object>” like when put into a pocket or pack or something

nightwingthedragon:
Putting people into pockets and packs would be considerably harder. I'd have to look at the code that deals with moving things into another thing's contents.

SpadesNeil:
Is that everything?

nightwingthedragon:
Aside from the possibility of changing things to use get_scooped, which is non-compulsory and can be patched in later. Code works just as well as it always did from what I've found with my testing.

code/modules/mob/living/carbon/carbon.dm
Leshana: I have a solution to improve this by making it generic instead of needing the "== 2", it is part of my general enhancement of the taur code to make it less hard-codey.
However it will be difficult for me to integrate properly until this PR is merged.  Can we expect it to be ready soon?

SpadesNeil: I'll merge it now.

Issue #187, opened 2015-10-20
Loyalty implants STILL not being applied at round start


SpadesNeil:
Only happens to HoP and Internal Affairs agent. This is not correct. Both should be implanted.

Leshana:
That's because the return statement is in the wrong place.

SpadesNeil:
oh derp

Leshana:
Is dis fixed?

Arokha:
This is fixed.

PR #192, opened 2015-10-22
Loyalty Implant For HOP

Leshana:
Remove spurrious return statement to fix Issue #187

### DIFF
--- a/code/game/jobs/job/captain.dm
+++ b/code/game/jobs/job/captain.dm
@@ -87,7 +87,6 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 			H.equip_to_slot_or_del(new /obj/item/weapon/storage/box/ids(H), slot_r_hand)
 		else
 			H.equip_to_slot_or_del(new /obj/item/weapon/storage/box/ids(H.back), slot_in_backpack)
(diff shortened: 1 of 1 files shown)


code/game/jobs/job/captain.dm
```diff
@@ -87,7 +87,6 @@ var/datum/announcement/minor/captain_announcement = new(do_newscast = 1)
 			H.equip_to_slot_or_del(new /obj/item/weapon/storage/box/ids(H), slot_r_hand)
 		else
 			H.equip_to_slot_or_del(new /obj/item/weapon/storage/box/ids(H.back), slot_in_backpack)
-		return 1
 
 		var/sound/announce_sound = (ticker.current_state <= GAME_STATE_SETTING_UP)? null : sound('sound/misc/notice2.ogg', volume=20)
 		captain_announcement.Announce("All hands, [H.real_name] is the Head of Personnel!", new_sound=announce_sound)
```
ShadowLarkens: This should probably be removed, it's almost certainly why the above return was there in the first place.

SpadesNeil:
Oh, I bet this is why internal affairs agents also have problems being implanted. Odd that I didn't notice this before.

Issue #87, opened 2015-10-06
Chaplain Job Titles, new code missing things, 


Molenar:
Molenar wrote:
we lost at least some of the alternate job titles on code swap. I noticed a lot less available under the chaplain options. Used to use 'Priestess' all the time. Haven't looked through all of them to find others that were lost.
Aces wrote:
Chaplain is one job that should get its alt titles back.

Other jobs, especially those in command, are sticking with what they have now and I'm not changing it back because it broke things and caused confusion.

But Chaplain can get its alt titles back. Submit an issue on Github.

HawkerTheGreat:
If you could give me some ideas for alt titles, I'll add them... Or for any other job. We should stick to not giving heads titles, though. I agree with Ace.

PR #194, opened 2015-10-22
Updated alt_titles Chaplain

HawkerTheGreat:
I looked at one of the issues in this github and decided to set on something simple to help out people, the ALT TITLES are the only variable I changed and I added quite a few, understand this is only the file for the chaplain. I will throw the changes down bellow since I haven't used github in a while and I'm unsure if it records any of that.

alt_titles = list("Counselor","Therapist","Preist","Cleric","Divine","Monk","Preistess","Pope")
As above, I added a few for example and might decide to add more if we really need it. I want to look at the other jobs and do the same if this commit gets accepted. - I've also tested this and compiled it.

### DIFF
--- a/code/game/jobs/job/civilian_chaplain.dm
+++ b/code/game/jobs/job/civilian_chaplain.dm
@@ -10,7 +10,7 @@
 	selection_color = "#dddddd"
 	access = list(access_morgue, access_chapel_office, access_crematorium, access_maint_tunnels)
 	minimal_access = list(access_morgue, access_chapel_office, access_crematorium)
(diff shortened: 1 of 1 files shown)


HawkerTheGreat:
https://github.com/SpadesNeil/VOREstation/issues/87

SpadesNeil:
Yank Pope out. That doesn't make any sense. Add Imam while you're at it. Rabbi too.

HawkerTheGreat:
Thanks Ace, I'll add that right now.

PR #200, opened 2015-10-23
Hackable Jukebox

Leshana:
* Made Jukeboxes hackable
* When hacked it will display an expanded list of available tracks.
* Other wires will shock the user, do special effects, or nothing.
* Pulsing wires gives hints, cutting does the effect


Arokha:
I approve.

Issue #169, opened 2015-10-19
Upgrading the destructive analyzer actually makes it worse!


Treasa729:
When the destructive analyzer is outfitted with better parts, it is supposed to be able to deconstruct lower reliability items, but when upgraded, it has the opposite effect! Things that it would be able to deconstruct without upgrades can no longer be deconstructed, even if they have the same reliability!

   I verified this myself by upgrading the analyzer with full reliability parts, (So that's not part of the issue) and tried to deconstruct an item with 90 reliability with no success. After re-outfitting it with the stock, weakest parts, it was able to deconstruct the same item, not remade or anything else!

   In addition, the analyzer, unupgraded puts 60% of the scanned material into the protolathe's material storage, wheras the upgraded analyzer gives none of the material back at all!

Arokha:
Probably a less-than greater-than math problem in the code.

SpadesNeil:
Oh. Wow. Bay never fixed this. I fixed this bug in the last code, told them, and then... ugh...

I'll look at it later... -_-

SpadesNeil:
I think the issue lies here.
[code]/obj/machinery/r_n_d/destructive_analyzer/RefreshParts()
	var/T = 0
	for(var/obj/item/weapon/stock_parts/S in src)
		T += S.rating
	decon_mod = T * 0.1
	min_reliability = 93 - T[/code]

Arokha:
The decon_mod line is definitely wrong. That's a multiplier for how many materials it gives back. It starts at 1.

Each part has a 'rating' and the default for the stock 3 parts is all 1, so you would think it should be a modifier of 1, sure.

But what it APPEARS they were TRYING to do is take the 1 as a base then sum up the ratings and multiply them by 0.1, to get like... 1.6 reliability if you put in three rating 2 parts. That... is not what that math does. This can be fixed by making that:
> decon_mod = (T * 0.1) + 1

But even then, simply taking out the starting parts and putting them back gives you 1.3, so I am not sure they thought this through.

I'll rewrite the whole section.

Arokha:
> min_reliability = 93 - T

This appears to make sense, though. I'm not sure why this would cause issues. I'll have to test it ingame.

Arokha:
I just realized that it's working exactly as intended, however they have it set to be 100% efficient at map load, and then the first time RefreshParts() is called, brings it back down to 30% like a reasonable thing.

[...]

PR #203, opened 2015-10-23
Machine Construction Fix

Arokha:
A cheap trick for #169
Don't worry, this makes it more consistent with other things, so it's not
like duct-tape and string or anything. This needed to be changed, too, but
doesn't fix the other weird issue.


Arokha:
Ok, finished a full machine construction fix to prevent this dupe parts issue.

PR #205, opened 2015-10-23
Civilian alt_titles added

HawkerTheGreat:
I wondered why they had removed some of the extra titles from the new version of baycode. Of course, these things only usually add to the game if we're talking about using them in a RP server. I decided to add a few back, but I'm still lacking on ideas for other job titles.


SpadesNeil:
Forcaster should be removed. It's not something that really makes sense in space. The rest looks excellent however.

HawkerTheGreat:
Done, I'm glad I can help in even the slightest ways.

PR #220, opened 2015-10-25
Makes all autoinjectors not microcillin

Arokha:
Yeaaaaaaaah whoops.


Michab02:
Nononono, leave it. This is perfect. D:

PR #236, opened 2015-10-27
Atmos Sensor Computers (Fix #234)

### LINKED ISSUE #234: Atmos Control Computers cannot be re-assembled
If you screwdriver an atmos control computer (Such as the sensors readout computer in toxins mixing room) and then screwdriver again to re-assemble, it no longer works. (Says no connected sensors)

Leshana:
* Circuit board needs to retain not only the frequency but also list of
sensor IDs in order to properly rebuild the same computer.
* You cannot just set the frequency var, you must call set_frequency()
for it to take effect
* While we are here, lets make examining the board show the frequency
configured.
* While we are here, lets make it possible to link up air sensors and
atmos computers!

### DIFF
--- a/code/game/objects/items/devices/multitool.dm
+++ b/code/game/objects/items/devices/multitool.dm
@@ -20,3 +20,8 @@
 
 	origin_tech = "magnets=1;engineering=1"
 	var/obj/machinery/telecomms/buffer // simple machine buffer for device linkage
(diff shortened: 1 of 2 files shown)


SpadesNeil:
No one has discussed this in 4 days. I think you can merge it safely.

PR #240, opened 2015-10-28
Make Micro/Macro Reagents Craftable

Leshana:
* Added "Normalcillin" a drug that returns you to normal size!
* Refactored the code for these reagents, removed dead code and cleaned
up the if statements to make it more datum driven and remove
duplication.
* Added chemistry recipes for crafting the micro/macro/normal cillin
* Added "sizeoxadone" as an intermediate precursor to them for game
balance.
* Added a fun effect upon mixing micro and macrocillin together.

### DIFF
--- a/VOREStation.dme
+++ b/VOREStation.dme
@@ -504,6 +504,7 @@
 #include "code\game\objects\effects\explosion_particles.dm"
 #include "code\game\objects\effects\gibs.dm"
 #include "code\game\objects\effects\glowshroom.dm"
+#include "code\game\objects\effects\grav_pull.dm"
 #include "code\game\objects\effects\landmarks.dm"
 #include "code\game\objects\effects\manifest.dm"
 #include "code\game\objects\effects\mines.dm"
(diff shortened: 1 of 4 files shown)


Leshana:
I've tested all this, but I'd like some input on the actual recipes for creating micro/macrocilian etc. I am pretty sure that the sizeoxadone precursor is a good plan (the idea was to make it non-trivial to make, since even just 1 unit is enough to shrink someone)
But I basically picked precursor chemicals that seemed interesting, there might be ones that would have better lore or balance.

OrbisAnima:
I would use something else instead of Doctor Delight, I believe that Phoron as a catalyst is good, but a barmans Drink is not so much... clonexodane might be nice, since it with bicardine and phoron is the precursor to the wonder drug that regenerates organs. I would go with clonex, phoron as catalyst and something else...ethanol maybe? or tramadol as you set.

SpadesNeil:
Remove this "do not merge" tag once you guys work out the recipe.

Leshana:
That seems reasonable enough. I have changed it to be clonexadone.   Everything else look good?

OrbisAnima:
I think it's perfect!

PR #241, opened 2015-10-28
New religious book cover and kimono

Arokha:
Both sprited by Molenar.


code/game/jobs/job/civilian_chaplain.dm
```diff
@@ -104,6 +104,9 @@
 							for(var/turf/T in A.contents)
 								if(T.icon_state == "carpetsymbol")
 									T.set_dir(10)
+					if("Shinto")
+						B.icon_state = "shinto"
+						B.item_state = "syringe_kit"
```
Leshana: Why is the item_state  "syringe_kit" ?

Arokha: I have no idea. Refer to the file. That is how they are all written in the stock switch statement. Just copied what the one that was there had.

Issue #168, opened 2015-10-19
Satchel clips through spiderbutt


AliciaJewel:
If you're a spidertaur with a satchel, the satchel will clip through your spider half if you're facing north as shown here: http://puu.sh/kPqRq/96e35a063f.png

nightwingthedragon:
Not sure this issue can be fixed since the tail layer is apparently below the satchel layer for... Some... Reason. I'll look into it at some point.

Arokha:
This is indeed what the issue is. You can swap the layers but then you end up with wings being rendered ontop of back items and other stuff. Basically you trade graphics glitches for other glitches. Maybe worth it though? Dunno. Could also tweak the spider sprites.

SpadesNeil:
plz fix looks horrible

SpadesNeil:
Was "fixed"

PR #244, opened 2015-10-28
Re-arranged update_icons layers

Arokha:
Moved tail layer down. Prevents it from rendering under back and suit
slot.

There are downsides to any layer placement. Nothing will make this perfect
unfortunately. Some placements decide whether something looks good from
the front or back, but not good from both.

And not every suit slot item is really a back item. Stun batons generally
aren't. But MOST are weapons worn on the back, and we can only make it
look good for MOST things, sadly.


Arokha:
"Fix" for #168

Leshana:
Either solution is a compromise; there isn't a solution which will fix all the glitches with this framework.   Lets evaluate if this ordering looks better in more cases.

PR #245, opened 2015-10-28
Walletchange

Arokha:
Fixed wallet sprite for CC ID, made it hold permits, reduced item count to 7, removed some potentially-OP items from being allowed.


nightwingthedragon:
Removing crayons? Seeds? Why?

Arokha:
Just removed stuff that people could use for doing jobs in general so it doesn’t increase the scope of what it’s for. Crayons may have been an accident. I’ll edit them back in.

Arokha:
See commit above.

nightwingthedragon:
And seeds? Really, seeds? There's nothing wrong with those fitting in a wallet. They're small and flat and you can fit dozens more into a plant bag. Seeds in wallets are completely fine.

PR #256, opened 2015-11-02
Added HUD Glasses to officer spawn

HawkerTheGreat:
As above.

### DIFF
--- a/code/game/jobs/job/security.dm
+++ b/code/game/jobs/job/security.dm
@@ -148,6 +148,7 @@
 		H.equip_to_slot_or_del(new /obj/item/clothing/under/rank/security(H), slot_w_uniform)
 		H.equip_to_slot_or_del(new /obj/item/clothing/shoes/jackboots(H), slot_shoes)
 		H.equip_to_slot_or_del(new /obj/item/device/pda/security(H), slot_belt)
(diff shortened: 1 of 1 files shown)


Leshana:
Can we confirm: I thought that you could select this item in custom items, and it gave it selectively only if you were in a security job. Is this true?

SpadesNeil:
Yea, I thought you could pick hud glasses in the custom items slot, too.

Arokha:
You can pick "Security HUD", aka the thing worse than HUDSunglasses.

SpadesNeil:
Well all of the security lockers spawn with hud sunglasses, and correct me if I am mistaken, but you also *don't* spawn with a flasher I think.

HawkerTheGreat:
Yeah, You can pick hudglasses - they come in the lockers as well as flashes which spawn in all of sec's pocket - but security all spawns with glasses except for the normal security guard. Detective, Warden, HoS... All spawn with hud glasses. I only assume it would be better for security officers to have the same.

HawkerTheGreat:
I'm also debating if flash and cuffs should be removed from spawn since they're in security... Technically since they take up inv space, they're clutter for people like me. You end up putting them in your belt anyways, but they spawn in lockers too.

Either way.

Leshana:
Okay that is good enough reason, if all the others do.

PR #260, opened 2015-11-03
simple_animal changes

Arokha:
Lots of changes.

-Gave simple_animal mobs their own 'simple' belly type that sends messages
to prey, in lieu of having an actual RP partner.

-Defined 48 or so emotes to be used for this purpose and added comment
blocks for people to add the rest that I wasn't creative enough to do.

-Rebalanced how fox and cat hunt. They were a bit 'homing missile' in the
past, they are a bit easier now.

-OOC escaping from a mob adds you to be excluded from being re-eaten for 5
seconds.

-Mobs now respect the no-digesting flag, and use the same vore bellies as
everyone else.

-Added custom emotes for Renault and Runtime to emote into their bellies
when they have prey. Go get ate!

-Refactored OOC escape code to clean up belly listings inside the pred and
just use the belly functions to get out.

-Split OOC escape message into two, depending on mob/PC escaping from, and
made admin message reflect this.

-Moved 'digestable' up to /mob/living so that practically everthing can
have that flag that's alive.

### DIFF
--- a/VOREStation.dme
+++ b/VOREStation.dme
@@ -1577,6 +1577,7 @@
 #include "code\modules\vore\belly\belly.dm"
 #include "code\modules\vore\belly\belly_balls.dm"
 #include "code\modules\vore\belly\belly_boobs.dm"
+#include "code\modules\vore\belly\belly_simple.dm"
 #include "code\modules\vore\belly\belly_stomach.dm"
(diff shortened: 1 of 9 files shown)


Leshana:
This looks like it should work to me.

PR #261, opened 2015-11-03
Mob digestion toggle verb

Arokha:
It's in the vore tab. Turns on digestion for 20 minutes, then back off.
Can also turn it off early. Has a warning box for both directions.

### DIFF
--- a/code/modules/vore/vore.dm
+++ b/code/modules/vore/vore.dm
@@ -46,7 +46,7 @@
 			spawn(12000) //12000=20 minutes
 				if(src)	insides.digest_mode = "Hold"
 	else
-		var/confirm = alert(usr, "This mob is currently set to digest all stomach contents. Click the button below to disable this.", "[name]'s Digestion", "Disable", "Cancel")
+		var/confirm = alert(usr, "This mob is currently set to digest all stomach contents. Click the the disable button below to disable this.", "[name]'s Digestion", "Disable", "Cancel")
 		if(confirm == "Disable")
 			insides.digest_mode = "Hold"
 
--- a/code/modules/vore/vore.dm
+++ b/code/modules/vore/vore.dm
@@ -3,6 +3,10 @@
 
 /mob/living
 	var/digestable = 1					//Can the mob be digested inside a belly?
+	var/datum/voretype/vorifice = null	// Default to no vore capability.
+
+	// TODO - Rename this! It is too conflicty with belly.internal_contents
+	var/list/internal_contents = list()
 
 /mob/living/simple_animal
 	var/isPredator = 0 					//Are they capable of performing and pre-defined vore actions for their species?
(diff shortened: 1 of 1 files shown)


code/modules/vore/vore.dm
```diff
@@ -13,26 +17,38 @@
 	var/list/prey_excludes = list()		//For excluding people from being eaten.
 
 	//We have some default emotes for mobs to do to their prey.
-	var/list/stomach_emotes = list("The insides knead at you gently for a moment.",
+	var/list/stomach_emotes = list(
+									"The insides knead at you gently for a moment.",
```
ShadowLarkens: Not an issue persay, but I'm not sure why this was actually changed.

code/modules/vore/vore.dm
```diff
@@ -13,26 +17,38 @@
...
+	set desc = "Enables digestion on this mob for 20 minutes."
+	set category = "Vore"
+	set src in oview(1)
+
+	if(insides.digest_mode == "Hold")
+		var/confirm = alert(usr, "Enabling digestion on [name] will cause it to digest all stomach contents. Using this to break OOC prefs is against the rules. Digestion will disable itself after 20 minutes.", "[name]'s Digestion", "Enable", "Cancel")
+		if(confirm == "Enable")
+			insides.digest_mode = "Digest"
+			spawn(12000) //12000=20 minutes
+				if(src)	insides.digest_mode = "Hold"
```
ShadowLarkens: Good on the sanity-checking... but I don't think this actually does anything, if src changes, I believe all procs called from it, including spawn()ed procs are killed.

code/modules/vore/vore.dm
```diff
@@ -13,26 +17,38 @@
...
+	else
+		var/confirm = alert(usr, "This mob is currently set to digest all stomach contents. Click the button below to disable this.", "[name]'s Digestion", "Disable", "Cancel")
```
ShadowLarkens: This could probably be phrased a bit better, it is kindof awkward with "Click the button to do X"

code/modules/vore/vore.dm
Leshana: This is true.

code/modules/vore/vore.dm
Arokha: Looks prettier in the code, easier to read because it lines up in the .dm file. Impossible to tell on github due to formatting.

code/modules/vore/vore.dm
```diff
@@ -3,6 +3,10 @@
 
 /mob/living
 	var/digestable = 1					//Can the mob be digested inside a belly?
+	var/datum/voretype/vorifice = null	// Default to no vore capability.
```
ShadowLarkens: This is really just turning into nitpicking, but this comment is out of line with the above comment.

code/modules/vore/vore.dm
ShadowLarkens: Ah, got it.

code/modules/vore/vore.dm
```diff
@@ -13,26 +17,38 @@
...
+	set src in oview(1)
+
+	if(insides.digest_mode == "Hold")
+		var/confirm = alert(usr, "Enabling digestion on [name] will cause it to digest all stomach contents. Using this to break OOC prefs is against the rules. Digestion will disable itself after 20 minutes.", "[name]'s Digestion", "Enable", "Cancel")
+		if(confirm == "Enable")
+			insides.digest_mode = "Digest"
+			spawn(12000) //12000=20 minutes
+				if(src)	insides.digest_mode = "Hold"
+	else
+		var/confirm = alert(usr, "This mob is currently set to digest all stomach contents. Click the the disable button below to disable this.", "[name]'s Digestion", "Disable", "Cancel")
```
Arokha: Updated the message.

code/modules/vore/vore.dm
Arokha: This is not the case in the .dm file. Again just a github formatting issue.

ShadowLarkens: oh right github uses indentation 8, DM uses 4.

Leshana:
This is proper then.

code/modules/vore/vore.dm
ShadowLarkens: >Click the the disable button below to disable this
>the the

Might look a tad bit better like this
```
alert(usr, "This mob currently digests it's stomach contents. Would you like to disable this?.", "Disable [name]'s Digestion", "Disable", "Cancel")
```

ShadowLarkens:
Since this is a minor feature that really can't harm anything, I'll merge this now before the patch. @SpadesNeil / @nightwingthedragon yell at me/Arokha later if that's bad.

Issue #263, opened 2015-11-03
Skrell are forced to be vegan!


nightwingthedragon:
This seems dumb on a station that centres around eating other people, but right now, by lore and by Baystation's coding, Skrell take toxin damage every time they eat anything that contains "animal protein", which is, as far as I've heard, mechanically identical to nutriment, but deals toxin damage to skrell.

Arokha:
See PR.

Arokha:
Now there are two different PRs to change this. XD

HawkerTheGreat:
Yeah... We were both attacking the same thing, See Arokha's ref for what I commented. Removing the tag causes problems.

Arokha:
Like what?

Arokha:
Actually Leshana had a good suggestion. Why not replace the stupid IS_SKRELL, IS_DIONA, etc with effect bitflags?

nightwingthedragon:
What's the word on this?

SpadesNeil:
Was this fixed also?

nightwingthedragon:
This needs looking at please.

Screemonster:
[06:16:22] Jemli: also Scree
[06:16:59] Jemli: in the chemical-regents.dm, there's this for animal proteins
[06:17:18] Jemli: on_mob_life(var/mob/living/M, var/alien)
    if(alien && alien == IS_SKRELL)
     M.adjustToxLoss(0.5)
     M.nutrition -= nutriment_factor
    ..()
[06:18:24] Jemli: also for the regent egg for the same reason
[06:18:40] Jemli: just take those out and it should change it so that they can eat meat
[06:21:04] Jemli: skrell also have a thing in the same file where they have any booze they drink multiplied by 5

PR #267, opened 2015-11-03
Change for #263

Arokha:
This change is to remove the IS_SKRELL reagent tag from the Skrell
species. If this is what you want to do, merge this PR.

Effectively, this makes reagents treat Skrell like humans.


HawkerTheGreat:
I was working on the same exact branch O.o
Woops. - Rather then remove the tag from it, which will break it... I removed the lines of code dealing dmg. The IS_SKRELL is linked to a few different files.

HawkerTheGreat:
I feel this will cause less problems -
https://github.com/SpadesNeil/VOREstation/pull/268

Arokha:
You could do it either way. My way won't cause problems, though. The reagent tags are just that, reagent tags.

I think NW's reasoning was that Skrell should be the same as other species, so removing their differentiation rather than removing the effects of it is what I was going for.

Honestly both PRs could be merged and it'd have the same effect.

PR #268, opened 2015-11-03
Remove Egg/Fat dmg for Skrell

HawkerTheGreat:
I removed the lines of code that damage skrell when they ingest fat or egg in-game. Without this, its not harmful to eat people.

https://github.com/SpadesNeil/VOREstation/issues/263
Upon request.


Arokha:
See my comment on #263 about changing this into bitflags for effects instead.

Issue #274, opened 2015-11-05
CC Rep/Obs/Adm HUD Icon


Arokha:
There's a nice new NT icon for people but not for our custom jobs. Need to apply the NT icon to these jobs for the hud instead of the dumb unknown questionmark.

SpadesNeil:
This cannot be fixed with how the system currently handles alt title jobs. The whole system needs to be gutted. Marking as wontfix, because this involves more than a single patch. It involves redoing the entire code of how hud glasses work.

Arokha:
I'm cool, so I'll fix it anyway without doing that.

SpadesNeil:
Nevermind I'm a doof.

Issue #271, opened 2015-11-04
Sagaru (sergal language) font color


silveryferret:
Can we please get a different font, or a different color, or /something/ to separate it from galcom? :x Just... anything. It's so confusing sometimes..

SpadesNeil:
It used to be red. I don't know why it's not red anymore. It should be red.

Issue #117, opened 2015-10-12
Emergency messaging.


Tasald:
Currently only the captain can send an emergency message to cent com, I tried as HoP and couldn't. I feel like all heads should have this access at least because there's not always a captian on board. I also think that the HoP and HoS should be able to raise and lower the status to blue, for the same reason as there's not always a Captain on board, granted a hop can just give them self's the accesses, so its more for the HoS

SpadesNeil:
I agreed with this notion.

Arokha:
I would agree with the HoS being able to do it to blue without talking to people. They are implanted, and would probably be the first to know of a 'credible threat to the station'. I think red should still require cards or a captain though.

The messaging though... why? Why not just send a fax? Every head already has a fax machine, and there's one on the bridge too (breakroom).

Tasald:
Red should require cards but it's in the wall card reader isn't it still?

For the messaging, it takes much less time, last time I checked there was no cool down on it either unlike the fax machine where you have to type it out onto a piece of paper, then put the paper in, then your id, then send it, and then either wait for five mins for it to re calibrate or go find another fax machine. During an emergency situation, there would be should be at least one head near or on the bridge, so if needed they could send an update to CC at will with out needing to run around the station looking for a fax machine if there was an important update. I know as HoS I typically stay on the bridge, its centralized and has both cameras crew monitor, and a fax machine which I admit, would be the first device I'd use to tell CC if I thought we might need help with something, but then the thing goes to hell faster then the fax machine can calibrate and I'm stuck wanting to be able to send them an emergency request right there with out having to run all the way back into sec, doging what ever's in the hall possibly, to get to the office and type out a message that I could have sent fast via the computer. This goes the same for any of th heads.

Arokha:
It looks like heads were ALWAYS supposed to be able to set the alert level, based on what I see in the file for this, but their check to see if you're a head isn't working.

Issue #63, opened 2015-10-02
Farm bots aren't buildable anymore


OrbisAnima:
gathered all the tools, wasn't able to assemble one

SpadesNeil:
Farmbot.dm has been turned back on. Closed issue for now. Will be in next push.

SpadesNeil:
Okay I lied. It pukes up a bunch of compile errors.

OrbisAnima:
I am not the guy to ask about Wiki :P

SpadesNeil:
Giving up on this for now.

PR #275, opened 2015-11-05
Various Patches

Arokha:
Fix for #271 
Added a CSS style for .sergal, light-blue.

Fix for #274 
Added the new CC titles to the already-hardcoded list of titles checked by the HUD.

Change for #117 
Removed the difference between heads and captains when checking for IDs on this console.

Work on #63 
Cleaned up some references to make them make sense in Bay, but still lots of work to do. Left it unchecked in .dme.

### DIFF
--- a/code/game/jobs/access.dm
+++ b/code/game/jobs/access.dm
@@ -427,7 +427,12 @@
 		"BlackOps Commander",
 		"Supreme Commander",
 		"Emergency Response Team",
-		"Emergency Response Team Leader")
+		"Emergency Response Team Leader",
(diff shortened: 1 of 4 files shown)


Leshana:
This seems fine if you ignore the farmbot.dm which is not included anyway.

ShadowLarkens:
Yeah, this looks okay, other than the whole "farmbot not actually going to compile at all" part, but I'm fairly sure it already doesn't compile.

Arokha:
Yeah, it doesn't. I just wanted to push work so people could continue it.

PR #276, opened 2015-11-05
Custom items

OrbisAnima:
Added Tsunderenyaa items.
(cultknife and casualsec)

Fixed science uniforms not rolling down


Leshana:
These look like approved items.

Issue #272, opened 2015-11-04
Micros as ID


Treasa729:
I think it would be awesome if it would be possible to use micros as ID for when opening airlocks, that way when you pick someone up, you don't necessarily have to take their ID off to be able to move around in their department and such! Is there some way to use whatever access the micro would have at the time (So what they'd have access to if they just walked at an airlock) and give that access to the macros holding them as well when macros hold them in hand? (or possibly in the ID slot :P) I think that'd be awesome for snatch styled stuff, heh.

Arokha:
This is done. Fixing some other issues before making the PR.

Issue #252, opened 2015-10-31
Issues with forcefeeding


Treasa729:
It seems that people can no longer force feed you once you reach a certain threshold of food, even if you'd still be able to eat by yourself! D:

SpadesNeil:
I thought I disabled that threshhold! Unfortunately I'm busy right now so I can't do any SS13 coding but I'm sure someone else will fix it.

Arokha:
This is done. Fixing some other things before making the PR.

Issue #99, opened 2015-10-08
Wiggle cooldown/randomness is much too sparse.


SpadesNeil:
I know we got the cooldown working but now we have the opposite problem as before.

Now it's so hard to make wiggle emotes appear that it's almost as if the feature doesn't exist.

Wiggle emotes should be more responsive, though a cooldown should still be in effect.

nightwingthedragon:
Whoever has the code for this, find out what it currently says, and think of a reasonable delay between squirms. Perhaps a random number between 3 and 5 seconds.

Arokha:
The cooldown is 2 seconds, and has a 40% chance of emoting. It appears that only stomachs have this feature, though. There's no struggle emotes for wombs or balls or whatever.

SpadesNeil:
Remove the % chance please? If it has a cooldown for 2 seconds then the 40% chance is redundant. That was just my hacky-ass fix for not knowing how to do cooldowns at the time.

Issue #42, opened 2015-09-30
Lysis-Isolation Centrifugue (Xeno Botany issue)


OrbisAnima:
If the Lysis machine had a disk when there is a power outage, it will take out the disk, but will not allow you to insert it again, or a new one, since it thinks there is a disk.

There is also the issue where if you try to destroy and then rebuild the machine there is no board for it.

the machine name is /obj/machinery/botany/extractor

Arokha:
I can't duplicate this. Turning off the power to the machines in the room doesn't seem to eject the disk, and it seems to let me insert/remove/insert before and after doing so.

OrbisAnima:
I couldn't duplicate it either, , but, there is still the thing where we should create the board for it

PR #277, opened 2015-11-05
Various Patches 2

Arokha:
Change for #272 
Updated micro holders to have a GetAccess() that the normal access-checking routines can call, that gets and returns the access from the micro inside the holder.

Fix for #252 
Feeding self had nutriment limit raised to 6000 but not feeding others. Corrected this.

Addition for #42 
Added two circuitboards to the printer for the xenoflora machines.

Change for #99 
Changed the chance of wiggling up to 60% from 40%, left cooldown at 2sec.


ShadowLarkens:
Looked over this, seems fine to me.

Leshana:
This all looks good.  Keep in mind deconstructing the flora machine for #42 will still not give a board back, but its now constructable!  Making it deconstructable can happen later.

SpadesNeil:
I was going to say remove the chance detection entirely. It no longer is required. The cooldown is enough.

PR #278, opened 2015-11-05
Make Xenobotany Machines Deconstructable

Leshana:
Previous update (Fixing #42) made these machines constructable.  Even
better that they be deconstructable as well!

To this end I added the parts to the contents in the constructors.
I also noticed it was using outdated code to handle opening the panel and deconstructing, so I upgraded to use the new machinery default methods.


Leshana:
Will @ShadowLarkens  or @nightwingthedragon or someone like that like this?

nightwingthedragon:
All tested and stuff, yeah?

Leshana:
Yeah

nightwingthedragon:
Good good

PR #281, opened 2015-11-06
Whitelist system for ears/tails

Arokha:
To prevent others from using character-specific ears/tails.

Also added all the existing people to be allowed to use theirs.

Also has my fluff ears/tail, since approval for that was dependant on
whitelist being added. So here it is!


Leshana:
Logical

PR #285, opened 2015-11-06
Wallet recode, additional color, fluff hypo

Arokha:
Three main changes:

1) Added a grey wallet, because the orange one is BLERGH.
2) Redid how wallets construct their icons to make it much easier to add
new wallet colors in the future.
3) My approved fluff hypo from the forums.


Leshana:
*nods sagely*

PR #286, opened 2015-11-06
Species equip_problems flag to help with job_controller et al.

### LINKED ISSUE #284: Akula don't start with sandals
Why not? Tajarans do.

Arokha:
job_controller.dm was hardcoded to give tajarans and unathi sandals and
nobody else. Now it uses equip_problems flags.

Also cleaned up the station species file in general. It was a bit chaotic.

Fixes #284


Leshana:
Good good, now the variables are in consistent order.

PR #295, opened 2015-11-09
Fix for micro holder in paw-grab

### LINKED ISSUE #292: Micros foot/pawgrab
It seems that when you pick up a micro by stepping on them with the grab intent barefoot, it doesn't make them into a proper micro, just the black silhouette rather than the micro you get now. You can't strip them, dress them up in hand, or use them as ID like when you pick them up with the help intent.

Arokha:
This rewrites the code for grabbing micros with your feet using the grab
intent. Now it uses the get_scooped proc on the mob instead of manually
constructing a holder.

Fixes #292


Leshana:
This seems like it should work to me.

PR #332, opened 2016-01-07
Threw bedsheets everywhere

Arokha:
Updated bedsheets dm file to include the colors we didn't have before.
Updated maps to use the new colors and whatnot.


SpadesNeil:
Map edits are to be passed through me. This resulted in a conflict and cannot be merged now.

Arokha:
You uh, left. And said you wanted to leave everything up to other people?

Also why are you committing straight to master?

SpadesNeil:
Because I'm an idiot and am too used to doing things my way that I forget I'm part of a team now. :U

But really though because of the nature of map commits, no two people can be working on the same map at the same time or they lose all of their progress, no matter how we do it. Map commits can be (and should always be) done directly for this reason if the changes are small, and I need to be informed about any changes to the Exodus maps.

PR #338, opened 2016-01-15
Adds urethral/bladder vore. 

KazeEspada:
You can now take people into your bladder and then pissify them if your into that. Otherwise, its just another fun place to carry people.


SpadesNeil:
This is an extreme fringe even among vore players and I have a feeling that its mere presence will bother a lot of people. For the same reason I haven't added anything scat related. I really don't think this is a good idea.

KazeEspada:
I could remove the watersports aspect of it. I added it for consistency.

SpadesNeil:
I still don't have good feelings about it. Why not just use the existing cockvore/ub mechanics? I could change the wording on it to make it more vague.

KazeEspada:
I feel like the option will be appreciated, and the worse bit is probably the message, but you could do that with emotes. I dont think its fair to compare this with scat.

nightwingthedragon:
TBH, the system was expanded for exactly this reason. Why not use the existing mechanics? Because we expanded the existing mechanics to be flexible, that's why. Honestly, people are generally pretty good about preferences on the station. I don't see what's wrong with adding the option to do this.

SpadesNeil:
But we're sticking them in holes that already exist, and bladder vore is so obscure that it's not even mentioned on vore sites.

Contrarily, tail vore would be fine to add. I'm not comfortable adding this.

nightwingthedragon:
Yeah, so like I said, change it to remove the watersports aspect. When we add custom bellies, people will be able to do things like this anyway. I may be speaking from a biased point of view, but I don't think this is a bad thing to be adding. First person who takes the initiative to make a new belly with the new code and instead of asking them to improve it, we gun it down straight away?

KazeEspada:
Well, yeah that was the one line I was iffy about adding. Would it fine if I didn't add a digest option at all?

KazeEspada:
Code is available, no longer interested.

PR #344, opened 2016-01-18
Rwalls will no longer be melted by fires.

KazeEspada:
I ded pls nerf.

Also, I personally don't think Rwalls should be breached by fire, it makes toxins unfun.

Points to consider:
Toxins fires render the entire department unusable.

Canisters can handle any temperature and any pressure up to several billion degrees.

Fire suits are rated to 30,000 degrees.

### DIFF
--- a/code/game/turfs/simulated/walls_reinforced.dm
+++ b/code/game/turfs/simulated/walls_reinforced.dm
@@ -6,7 +6,7 @@
 	density = 1
 
 	damage_cap = 800
-	max_temperature = 6000
+	max_temperature = INFINITY
 
 	walltype = "rwall"
 


Screemonster:
git gud.

Skill isn't about making biggest most burny fuckhot fire there is, it's about control.
Besides, I've seen people hit the cap on the current code without melting toxins. The burn chamber literally has a vent to space in it as well that should easily extinguish any fires provided the crew aren't running the hull shields for no reason again.

KazeEspada:
Or i could use a canister fire instead for all the safety and twice the temperature. Molenar made sure to tell me how useless bomb making was by bragging about two bluespace beakers and potassium and water making an easier explosion.  My death was only because i thought i wouldnt need cheap tactics to make a bomb.

JoanRisu:
My two cents: Following common sense does not always equal good game design (or story design, etc.) If it were so, The vault, armory, and bridge would be next to impossible to break into, the solars and the engine would be automatically switched on, R&D would be always completed. That would give security less incentive to be on the look out, engineers nothing to do until something goes wrong, and R&D would be rendered useless as a job. Common sense would have seen Voyager with a holographic security team, putting it in less danger, but the writers didn't put that in despite showing it to be possible.

I get where you are coming from, the mixing chamber should be rated higher in terms of temperature. That being said, It does seem like this isn't just a matter of "safety." It sounds like this is coming from the fact that you lost control of a mix resulting in a fire and now you don't want to get burned again. You've also argued that this shouldn't be a chore. I agree, a game should not be a chore but that doesn't mean it should also be easy and without consequence.

Frankly speaking, while this is an RP server (with vorish themes), safety is boring. Its why we have lasers and murderous clowns. 

I'm leaving this up to @SpadesNeil. Partly because of the thoughts above and partly because you changed a value to "Infinity" instead of asking for a map adjustment.

SpadesNeil:
Yea no. That makes toxins no fun and no risk. Besides, if it explodes, it gets immediately vented into space and gives the engineers something more to do. Previously people would commonly superheat some plasma and leave the room like that for the whole shift.

Now they can't do that.

PR #352, opened 2016-02-07
Fixes heart stopping chemical

killer653:
Changes "potassium_phorochloride" (A nonexistant chemical) to "potassium_chlorophoride", which is a chemical that exists and is meant to stop the heart.


killer653:
Ack, accidentally merged the skeleton race in with here. Going to fix this up real quick

PR #353, opened 2016-02-07
Fixes a non-existant chemical stopping hearts

killer653:
Changes an non-existant chemical to an actual chemical in the code.


SpadesNeil:
Curious, how is this chemical produced, then?

antsnap:
https://wiki.baystation12.net/Guide_to_Chemistry

The actual chemical that can be made is listed below from the wiki. I never heard of anyone making it though, but I guess it exists.

According to the chemistry guide.
 Potassium Chlorophoride 	Formula: 1 part Potassium Chloride, 1 part Phoron, 1 part Chloral Hydrate
Effects:
A specific chemical based on Potassium Chloride to stop the heart for surgery. Not safe to eat!

Cameron653:
You produce the chemical via:
("potassium_chloride" = 1, "phoron" = 1, "chloralhydrate" = 1), which results in 4 potassium_chlorophoride

potassium_chlorophoride deals losebreath, which causes suffocation up to 10 losebreath ticks IIRC, does 2 oxygen damage, and weakens (Basically a stun) them. It's also meant to stop the heart, but as the name was incorrectly put into the heartstoppers list, it didn't do its effect of stopping their heart.

PR #354, opened 2016-02-07
Adds a skeleton race for SilverTalismen

Cameron653:
Sprites are already in the game, under "icons/mob/human_races/r_skeleton.dmi" so sprites aren't needed for this. Tested and made sure this works.

PROS:
+Cold immunity
+Increased heat resistance
+Immune to toxins
+No need to breathe
+Has no blood.

NEUTRAL:
/Has no special attack.
/Feels pain
/Is immune to cold, but will still take brute damage from the vacuum of space.

CONS:
-Unclonable
-Take 1.5x more burn damage 
-Take 1.5x more brute damage

### DIFF
--- a/code/modules/mob/living/carbon/human/species/station/station.dm
+++ b/code/modules/mob/living/carbon/human/species/station/station.dm
@@ -336,7 +336,7 @@
     heat_level_2 = 1100
     heat_level_3 = 2200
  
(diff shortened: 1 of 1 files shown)


Cameron653:
Added in a quick fix for tajaran taking excessive brute damage. Changed brute_mod = 1.2 to burn mod = 1.2

SpadesNeil:
This new race is unlocked via whitelist, correct?

SilverTalismen:
Yes it should be locked via whitelist

PR #355, opened 2016-02-08
PDA in ear slot.

antsnap:
I hope I'm doing this git thing right. Made PDAs able to fit in the ear slot by reducing their weight class to tiny. IDs could fit in the ear for awhile. This makes RP scenes super convenient. 

Thread: http://forum.vore-station.net/viewtopic.php?f=20&t=458&p=2742#p2742


antsnap:
nvm i messed up :x

PR #356, opened 2016-02-08
PDA able to fit in ear slot

antsnap:
Lowered PDA's weight class to tiny, making it able to fit in the ear slot like it did with the old code. Is probably implantable now too, which also makes sense.

Thread: http://forum.vore-station.net/viewtopic.php?f=20&t=458

### DIFF
--- a/code/game/objects/items/devices/PDA/PDA.dm
+++ b/code/game/objects/items/devices/PDA/PDA.dm
@@ -9,7 +9,7 @@ var/global/list/obj/item/device/pda/PDAs = list()
 	icon = 'icons/obj/pda.dmi'
 	icon_state = "pda"
 	item_state = "electronic"
-	w_class = 2.0
+	w_class = 1.0
 	slot_flags = SLOT_ID | SLOT_BELT
 
 	//Main variables


antsnap:
Reason for this is mentioned in the thread, but it is a huge quality of life improvement to naked characters (small micemen, slime girls :3) and it is also very beneficial to heads while scene-ing, so they can more quickly check their beeping PDA to see what part of their department is on fire.

SpadesNeil:
We noticed this change when we came to the new code, and originally chose not to revert it because it's more realistic. Yet you present some reasons to bring it back anyway, so let's discuss it.

SpadesNeil:
We won't merge this until after a few posts.

antsnap:
Sure thing, but discuss where? The thread or here? Also, what do you mean by posts?

My points are made in the thread, so I'll briefly restate. It is a huge quality of life improvement and, when considering the other things in this game, isn't that much of a realism leap. Considering the frequency of nudity on the server, it makes sense to let pda's go on the ears. I can understand why this isn't needed on other servers, but from a pragmatist viewpoint it should go here.

Also you can deal with the 'realism' factor by pretending the PDA is foldable or something.

SpadesNeil:
Discuss it here. By posts I mean comments here. Also it's probably better not to question the realism as the PDAs aren't exactly fordable looking.

nightwingthedragon:
I never knew this was removed, and support it being re-added!

antsnap:
Thanks for the comments guys.

My argument is that the value of convenience offered from this feature is greater than the value of realism detracted by it. These kind of things are always subjective, but I think it would be a relatively minor detraction from total immersion.

Yeah I agree PDA's dont seem foldable, but I was just trying to conjecture a way to rationalize like we do with SSD and afk.

Cameron653:
It'd be pretty useful for people that aren't wearing clothing.

Issue #309, opened 2015-12-08
Change agressive grab when cuffing to normal grab


Dhaeleena:
Having to grab and upgrade the grab to cuff is really bad, the only way this can be done is if you leave the person in complete pain so they cant stand up in a long while, flashing and flashbangs doesn't work because the person easily stands up before you can even upgrade the grab. Having to just use normal grab to cuff would be a lot better and less annoying

SpadesNeil:
Fixed in next server update.

PR #367, opened 2016-02-10
Handcuffs grab fix

killer653:
Makes it so handcuffs only need a passive grab, as opposed to a aggressive  grab.

Fixes issue #309


SpadesNeil:
Good 'cuz this was fucking stupid. :D

PR #368, opened 2016-02-11
Slime People Race!

antsnap:
Note that atm in the code, they are slightly different and certainly distinct from xenobio slime people. Making them the same could be complicated (name issues etc.), so lets make that a goal for later.

PROS:
+No slip (same as xenobio slime people)
+No breathing
+No pain
+Slime glomp special attack 
	(same as xenobio slime people)
	not sure what this is, or if its better than normal attack, but for consistency I didn't touch it.
+Slightly improved night vision? 
	(Humans default with 2 I think, xenobio slime people had this set to 3 so I copied that)
+Xenobio slimes won't try to eat you.

NEUTRAL:
/Special feels for hot and cold
/Different from xenobio slimified people (no poison immunity as they have, but non-xenobio have more benefits overall e.g. pain immunity)
	done mostly for compatibility. Lore wise xenobio slimified people are more 'unstable' and don't reap as many rewards
/No ventcrawling (yet anyways)
/HAIRS!
/No facial hair though.
/No unique language yet (use Skrell as your learned secondary for immersion)
/No visible genitals/nipples (RP forming them during scenes if you'd like)
	Controversial, but allows naked slime chars without security having to freak
/Has blood (bluish color, not sure how to do dynamic blood color yet)
/Slime core organ!

CONS:
-Unclonable
-Take 1.2 Burn Damage 
	(20% more, not that big. meant to balance their immunities out. Lore wise slime cells are more susceptible to temperature fluctuations in terms of metabolic functions, or something :3)
-Very weak to cold (distance between coldness severity is lower)
-Reach level 3 warmness slightly faster (at 800k vs 1000k, same as Tajaran)

Check out this forum thread for discussion on even more new ideas for this race. For now though, I think it's safe and ready to commit.


antsnap:
DONT MERGE.

I'm removing the NO_PAIN flag, since neither the skeleton nor paradise nor xenobio slimes had it.

Issue #348, opened 2016-01-19
Belly flavourtexts are saved per-player instead of per-character.


Screemonster:
When adding flavortexts to inside spaces and saving them, they appear to be saving the same text for every character the player has, rather than saving them in each specific character's file.

Steps to reproduce:
Go into a character
edit flavortexts
save slot
load another slot
look at flavortexts. the ones just entered will be there. Editing them and going back to the first character will overwrite them again.

SpadesNeil:
@nightwingthedragon , can you fix that please?

nightwingthedragon:
I honestly don't know what's causing that. I have the code set up just like the old code. I might need @Leshana to help.

nightwingthedragon:
Come next sync we can hopefully close this

SpadesNeil:
Marking as resolved. Pending confirmation.

Cameron653:
I believe this should be closed. To my knowledge, this is fixed.

SpadesNeil:
I thought I closed this.

PR #369, opened 2016-02-11
Makes inside flavor text save. V2

killer653:
This fix enables inside flavor text to actually save. Fully tested and working. 
No bugs were present during the test. Fixes issue #348 

Reposting this since github decided it'd be a /great/ idea to automatically switch a branch I was working on to this branch, so it totally screwed it up. Fixed now, and it shouldn't happen again.


nightwingthedragon:
Approoooooov'd o3o

PR #370, opened 2016-02-11
Slime race is ready! (Did quick fix and blance)

antsnap:
Small fixes and balance:

The special slime attack did no damage! Now it does normal damage, and has special text.

 In order to discourage this race becoming whitelist only, I removed the NO_PAIN flag. 
Slimes on other servers don't have NO_PAIN either so it didn't make much sense (thanks Slayerhero90)

SUMMARY:

Note that atm in the code, they are slightly different and certainly distinct from xenobio slime people. Making them the same could be complicated (name issues etc.), so lets make that a goal for later.

PROS:
+Not whitelisted; Slime People for all!
+No slip (same as xenobio slime people)
+No breathing
+Slime glomp special attack 
	(same as xenobio slime people)
	does normal damage with special text, but has potential to be expanded
+Slightly improved night vision? 
	(Humans default with 2 I think, xenobio slime people had this set to 3 so I copied that)

NEUTRAL:
/Special feels for hot and cold
/Different from xenobio slimified people (no poison immunity as they have, but non-xenobio have more benefits overall e.g. pain immunity)
	done mostly for compatibility. Lore wise xenobio slimified people are more 'unstable' and don't reap as many rewards
/No ventcrawling
/HAIRS!
/Skin colors!
/Eye colors!
/No facial hair though.
/No unique language yet (use Skrell as your learned secondary for immersion)
/No visible genitals/nipples (RP forming them during scenes if you'd like)
	Controversial, but allows naked slime chars without security having to freak
/Has blood (bluish color, not sure how to do dynamic blood color yet)
/Slime core organ!
/Special gibs!

CONS:
-Unclonable
-Take 1.2 Burn Damage 
	(20% more, not that big. meant to balance their immunities out. Lore wise slime cells are more susceptible to temperature fluctuations in terms of metabolic functions, or something :3)
-Very weak to cold (distance between coldness severity is lower)
-Reach level 3 warmness slightly faster (at 800k vs 1000k, same as Tajaran)

END SUMMARY

Thread for ideas and discussion here. I will also respond on github of course~


antsnap:
Small edit was made. Turns out the feature of being auto-friendly with xenobio doesn't work yet so I removed it from the summary.

SpadesNeil:
Is this race whitelist only or no? I say it shouldn't be whitelist only.

antsnap:
No, not whitelist only. Sorry forgot to mention. I want this race for all

PR #371, opened 2016-02-11
Fixed typo in preferences_savefil.dm. Made slime race clonable.

antsnap:
Was preventing successful compilation. Missing quotation mark.

Also removed the NO_SCAN flag from the merged slime race. By popular demand.

http://forum.vore-station.net/viewtopic.php?f=20&t=470&p=2872#p2872

Intent is to replace it later with a new revivial mechanic for the slime people (phoron injection into core?)

PR #374, opened 2016-02-12
Made slime race clonable.

antsnap:
Removed the NO_SCAN flag by popular demand.

In the future we will remove it, but give slime people another means of IC revivial (e.g. injecting core, xenobio cloner). However, those more preferable options will take time to code so, this is just a hold-over.


SpadesNeil:
Wasn't this handled by https://github.com/SpadesNeil/VOREstation/pull/371 ?

SpadesNeil:
Okay apparently not.

Issue #390, opened 2016-02-20
Black tail missing from tail sprite menu.


SpadesNeil:
Kisuke wants it back.

SpadesNeil:
@Arokha or someone feex plz

PR #392, opened 2016-02-20
Adds black wolf tail sprite back.

Cameron653:
Adds the black wolf tail sprite back. Fixes issue #390 
Image: http://puu.sh/nf6me/16c8ffd6db.png


SpadesNeil:
He meant cat ear but this is nice too!

Cameron653:
Added in with #393 to prevent any possible merge errors with the two.

SpadesNeil:
Did this add in the cat tail, also?

Cameron653:
No. I couldn't find the cat tail, only the wolf tail.

SpadesNeil:
It's just the same sprite used for the kitty ears. I think it's just in the wrong file.

Issue #385, opened 2016-02-19
Custom ear/tail menu doesn't list standard options for other species


Screemonster:
At present, the ear/tail menus allow you to have your species default, none, or one of the options from the fluff tails menu. There is presently no option to pick the default tail for a DIFFERENT species (such as giving a tajara the sergal tail sprite)

Adding the other default tails will allow for things like giving a sergal a unathi tail to make an anthro kangaroo, and so on and so forth.

SpadesNeil:
This is something I would want actually.

Screemonster:
I can definitely see people wanting to put tajaratails on birds when I finally get them finished so they can make anthro griffins.

OrbisAnima:
take the tail sprites from those species and add it to the list of tails made for the code, that's pretty much what I am doing for the Eros server. You can of course, wait until I finish with it and port it

PR #393, opened 2016-02-20
Allows species specific tails to be chosen.

killer653:
Allows any race to have another races tail (Ex: Tajaran with Sergal tail.) ALSO incorperates #392 
Adresses issue #385


Cameron653:
Sprites:
http://puu.sh/nf8X6/2339ed2b4c.png

SpadesNeil:
So is the black-only cat tail also included here? Because #392 only had the wolf tail. This was a request from Kisuke.

Cameron653:
This one has the black wolf tail, but not the black-only cat tail. I couldn't find the black cat tail anywhere in the icons. I might've not been looking hard enough, but I didn't see it anywhere in there.

SpadesNeil:
It might have to be chopped out of the cat ears sprite. I'll worry about this later.

PR #395, opened 2016-02-22
Adds nudity permit (jumpsuit) and mankini.

Cameron653:
Adds a nudity permit (select able "jumpsuit" that allows players to have all the abilities of a normal jumpsuit, but without the jumpsuit sprite.) and also a mankini, which used to be only obtainable by hacking the clothing vendor.


antsnap:
Oooh I wonder where the pockets are :3

PR #403, opened 2016-02-24
Adds "Spiders Parlor" and "The Space Bar" barsigns, and 7 extra posters.

Screemonster:
Added a bar sign that IPU sprited up to make things a little more spidery. Also added an awful pun that I could not resist.
Also I totally did not know that the bartender could swipe his ID on the bar sign to change it but apparently that's a thing.


Screemonster:
Also figured I might as well add the posters I made a while back while I'm at it.

Corrects a little error in the lamarr poster's description, and also adds 7 additional posters.

antsnap:
Step into my parlor said the spider to the fly :3

PR #406, opened 2016-02-25
Completely fixes womb TF. (No taur fix button needed.)

Cameron653:
Completely fixes womb transformations.  
Automatically updates the taur half, so the hotfix button isn't needed anymore.

Also makes it so the "change species" actually changes the prey's species and taur.

### DIFF
--- a/code/modules/vore/belly/belly_womb.dm
+++ b/code/modules/vore/belly/belly_womb.dm
@@ -74,15 +74,15 @@
 					var/TFchance = prob(10)
 					if(TFchance == 1)
 						var/TFmodify = rand(1,3)
-						if(TFmodify == 1 && P.r_eyes != O.r_eyes && P.g_eyes != O.g_eyes && P.b_eyes != O.b_eyes)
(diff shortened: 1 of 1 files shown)


SpadesNeil:
Bit of a hacky hotfix. What we really need to do is fix the taur sprites on update_icons while resized.

Cameron653:
I mainly did the first thing that came to mind. Let me see if I can get "P.update_icons()" to work for the tf.

Cameron653:
Testing a possible fix

Cameron653:
Made a fix that makes hair color, species, and eye color automatically occur. Really can't make the taur automatically appear without the taur fix button, as it's out of my knowledge. What is here currently is about as good as I can get it.

Cameron653:
Made a fix, taur halves now automatically show up, now hotfix button needed.

Cameron653:
Title changed to clarify that it's completely automatic, and doesn't require the prey to update their icon by using the fix taur button anymore.

Issue #414, opened 2016-02-28
Disable self harvesting for walking mushrooms made in Xenoflora


SpadesNeil:
It causes a cascade effect that results in the server crashing due to a swarm of exponentially reproducing mushrooms.

It's fucking hilarious though.

PR #415, opened 2016-02-28
Stops mushrooms from self harvesting (Hotfix)

Cameron653:
Stops walking mushrooms from self harvesting to fix issue #414


SpadesNeil:
Wow it was really that easy to change.

PR #416, opened 2016-02-28
Adds xenoarcheology cyborgs.

Cameron653:
Adds a xenoarch cyborg that is able to aide in xenoarcheology. 
It is meant to only help, and not be able to do the entirety of xenoarcheology and anomaly research.

**Features**:
Has all needed tools to do xenoarcheology excavation.
Is able to mine out anomalies with no errors, and no required human interaction.
Uses the xenoarcheology specialized drill, instead of the normal xenoarcheology pick set.
Has science radio comms.

**Comments**:
Requires a human to do anomaly research
Requires a human to initially swipe their ID on the suspension generator if the cyborg will be going after artifacts, and not just anomalies.

**Notes**:
Uses the sciency looking mining cyborg sprite and or the mining drone sprite. Depends on which one the user desires/selects.

### DIFF
--- a/code/global.dm
+++ b/code/global.dm
@@ -228,7 +228,8 @@ var/list/cheartstopper = list("potassium_chloride")                       // Thi
 // Used by robots and robot preferences.
 var/list/robot_module_types = list(
 	"Standard", "Engineering", "Construction", "Surgeon",  "Crisis",
-	"Miner",    "Janitor",     "Service",      "Clerical", "Security"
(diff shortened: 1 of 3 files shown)


SpadesNeil:
Oooooo... neat. Any icons for it yet or no?

killer653:
Working ATM, but I tested it earlier yesterday and everything was working. 

The icons it uses for the sprite is the advanced mining droid, and also the mining drone sprite. The advanced mining Droid sprite works well as it's meant for the science departmentas.

PR #430, opened 2016-03-06
Gives the service borg and combat borg two new sprites.

killer653:
**Additions**:
Gives the service borg and the combat (syndicate) borg one new sprite each.

**Syndicate borg**: http://puu.sh/nwX9o/19c4f4680e.png
**Service borg**:     http://puu.sh/nwXeh/46eec0a97f.png 

The syndicate borg fits right in, but the service borg was slightly finicky, but seeing as it's sprite is vastly different from the normal borg (It has an apron/maids outfit on, fitting the service borg style of being a helpful part of the crew), and also has no actual xenomorph tail, making it easy enough to tell the difference upon first glance.


killer653:
Closing this, as I'm going to do an all in 1 sprite addition set.

PR #433, opened 2016-03-07
Adds restriction to creating web bindings. (Urgent HOTFIX)

killer653:
Adds a nutrition restriction to web bindings.
Nutrition restriction is 30 units needed. It isn't a lot, but can be changed to be higher if need be.

Without the restriction, you could easily spawn in thousands of web tiles, causing massive lag.

Note: Also fixes a typo in tail vore.


antsnap:
yeah do this

PR #442, opened 2016-03-10
Demon language, birdsong language color fix. Also swimsuits

antsnap:
Added demonish language, fixed birdsong language color. Added swimsuits to custom loadout and the univend (purple is best :3).

The birdsong language was supposed to have a pinkish color when you type, but you have to define the colors in the css stylesheet, so I did. I think the birdsong color is too bright, but it's not my place to change it.

The demon language uses .8 , no other language was using it so shouldn't be an issue. Has a darkish redish color to it. It is selectable as a secondary for all languages, and is meant for succubus and other demon chars :3.


SpadesNeil:
Isn't "Demonic" a better name? Or better yet "Daemon"?

antsnap:
Oki doki

PR #445, opened 2016-03-11
Fix demon language color URGENT

antsnap:
When I changed the name I forgot to change the stylesheet color name to match :3


antsnap:
How do I mark this with urgent, because the language.dm won't compile without it probably :3

SpadesNeil:
For future reference, compile before each pull request in future.

SpadesNeil:
Also there's conflicts currently.

antsnap:
ok fixed now

Issue #404, opened 2016-02-24
Make shoeless people run at same speed shoed people do?


antsnap:
Would be good for those who RP barefoot chars.

Thoughts?

OrbisAnima:
just take off shoes speed... also, take off the tajaran boost to speed if you are going to do it. Some sec players will whine, but already having glass in the floor stunning and damaging you is punishment enough... and this idea might even make handcuffs on the orange shoes (legcuffs) see more use.

Cameron653:
It'd probably be a good feature, as I don't see too many situations in which this could be a bad thing, except for security. And even then, the prisoners should usually have shoes on anyways.

antsnap:
Consensus seems that's a good idea. I'd like to add we have the mechanic were you can put people in the foot slot as a macro if you walk over them with no shoes; by making barefootedness less annoying, hopefully that sees more use~

Ill try to have a PR today.

antsnap:
https://github.com/SpadesNeil/VOREstation/pull/446

PR #446, opened 2016-03-11
Being barefoot no longer slows. Lowered Taj bonus movespeed (reasons inside)

antsnap:
https://github.com/SpadesNeil/VOREstation/issues/404

So shoes actually applied a negative slowdown (aka a speedup). I removed this shoe speedup in order to make barefooted people move the same as shoed. However, this would make everyone slower than they currently are, so I had to change default species slowdown to the negative slowdown that shoes used to give.

Tajarans previously had a slowdown of -1, which would be equivalent to -2 now. However, the reason for this was (probably) because many Taj characters would also be barefoot cause paws and stuff. Since that reason is obsolete now that barefoot chars are no longer slowed, I lowered their speedboost by half.

I'd be in favor or removing taj speedboost all together for balance (so everyone has equal speed now). But, people might complain, and I guess it makes sense for cat people to move quicker cause, cats.

### DIFF
--- a/code/modules/mob/living/carbon/human/species/species.dm
+++ b/code/modules/mob/living/carbon/human/species/species.dm
@@ -90,7 +90,7 @@
 	var/siemens_coefficient = 1   // The lower, the thicker the skin and better the insulation.
 	var/darksight = 2             // Native darksight distance.
 	var/flags = 0                 // Various specific features.
-	var/slowdown = 0              // Passive movement speed malus (or boost, if negative)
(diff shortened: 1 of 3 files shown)


SpadesNeil:
I'd rather not have Tajaran be quicker than everyone else. I'd rather everyone be a little quicker instead and save the speedboost for a different species.

antsnap:
They were already quicker, but I'm glad you agree. Ill make their speed the same as the others

Issue #444, opened 2016-03-11
UB Oviposition mechanic


SpadesNeil:
Requested by a number of players.

The option to release a character as an egg if you're a Bird or Unathi would be nice.

Eggs should have a look and description based on the species it came from--including custom species.

"That is an Unathi egg. Some species of Unathi apparently lay soft-shelled eggs!"
"That is a Nevarean egg. Most Nevareans lay hard-shelled eggs!"
"That is a Human egg. Some humans--wait, what?"
"That is a Tajaran egg. Apparently that's what a Tajaran egg looks like. Weird."
"That is a Skrell egg. It's soft and squishy."
"That is a [custom_species] egg. It has a very unique look to it."

Player should be able to break out of the egg at any time--or someone else break it.

antsnap:
That's cute.

Shouldn't be too tricky. Someone make sprites~

killer653:
This would be a really nice feature, I'm going to look into it and see if I can get a prototype working at least

PR #448, opened 2016-03-11
Allows you to release someone in an egg after TF

killer653:
Allows you to transform someone into your species, just like the change species transformation, but it also encases them in an egg.

**Eggs are as shown**: http://puu.sh/nD4ot/3eecd990e5.png
Open and closed states.

**Notes**:

**It instantly transforms and adds them to the egg;** due to a bug that would cause eggs to spawn infinitely. This was the only way I saw to get around it.

**[BUG]** - Creates two eggs instead of one. Only one of the eggs holds the prey.

**[BUG]** - Grille hitting sound plays when the prey is resisting out of the egg. I tried to change it, but I wasn't able to fix it. Shouldn't be too big of an issue, as the predator can just take the prey out of the egg easy enough.

**The egg can be torn open by clicking on it** with an id on yourself (Any access works.)

**The prey can resist out of the egg to break out of it.** This process takes two minutes

Fully tested and working, and all notable features/bugs have been listed above. No other conflicts but the ones listed were found.

Partially addresses #444


killer653:
Adding something, don't merge

Issue #452, opened 2016-03-12
Birdsong language color is bright and hard to read for me.


antsnap:
Anyone else think so? It's a quick change. Right now the color is FFBBFF,which is very bright pink (really close to white, on a white background...)

I was thinking making it darker, like #e600e6.

You can compare different shades by plugging FFBBFF into http://www.w3schools.com/colors/colors_picker.asp

SpadesNeil:
#cc9900 Try this. I think this looks better, and fits the birdy mood.

Cameron653:
It'd be a easy enough fix. Would you still like it to be "#CC9900"?

SpadesNeil:
Go for it.

Cameron653:
This should probably be closed since a PR was merged that fixed this.

PR #465, opened 2016-03-17
Fixes birdsong color

Cameron653:
Makes the birdsong color #CC9900
Directly fixes issue #452


Cameron653:
Err, something seems to be wrong. Closing this untiil further notice.

Issue #461, opened 2016-03-14
Abandoned crates should explode when shot


SpadesNeil:
Brought up in #460

Players abuse emitters and lasers instead of opening the crates like they should, which leads to other abusive problems down the line when they start lasering open gun crates and stuff.

Screemonster:
At the very least, using a gun/emitter to bypass the puzzle entirely shouldn't be _objectively better in every way_ than solving the puzzle.
Not only is it faster and simpler, it's a guaranteed success.
I'll take a look at the crate code - would we want it to _always_ explode when shot, or have a small chance of successfully opening it?
(for reference on the odds, old crate had 30% chance by random guessing, while new crates have a 1/504 chance. With a tool, old crates went up to 70%, while a skilled player can crack the new ones pretty reliably - new ones are _way_ more skill than RNG in getting them open)

SpadesNeil:
I want it to always explode. You're shooting a supposedly tamper-proof crate packed with explosives to destroy its contents if tampered with.

Screemonster:
Coolcool. Same for emagging/eswording/emping?

SpadesNeil:
Yes.

Screemonster:
Makes sense, it _is_ a "cryptographic sequencer" so codebreaking should be its _thing_.

Line 167-171 of /code/modules/mining/abandonedcrates.dm have the boom - abandonedcrates.dm also has them unlock when emagged.

The code for _crates_ has a thing for being attacked with an emag _or_ an esword that is slightly different to what abandoned crates do. I'm not super-familiar with how byond-code handles inherited stuff, but does this mean that since abandoned crates are a child of crates, their version of what happens when you emag them takes precedence?

This is probably the first actual codey-thing I've tried to tackle :p

SpadesNeil:
I'll try to find it. It's something like bullet_hit() or bullet_act() or something along those lines, and it's part of the crate object class overall. I have to overwrite the function for abandoned crates so they explode.

Screemonster:
Cool. I flung a thing to you in skype but I think that was the explosion one.

One solution I was considering was giving crates a new var like "tamperproof" or "secure" or something like that and having them check that var when you shoot/bomb/emp/sword them. If 0, proceed as before. If 1, boom.

SpadesNeil:
There's a better way of doing that. I can just overwrite the function instead of adding new vars. Unless you intend on adding this to other crates.

Screemonster:
[...]

PR #469, opened 2016-03-18
Secure crates can be made tamper proof now

SpadesNeil:
This will deter players from abusing projectiles to open boxes, as there
is now a chance to destroy the contents and cause a (small) fire, or a
small explosion that will alert security.

Crates will not detonate if unlocked so the CE doesn't have their life
made hell for using an emitter to unlock the supermatter core.


SpadesNeil:
Addresses #461

Screemonster:
No change to make 'em boom on eswording or emping? Admittedly those things are gonna be pretty rare.

SpadesNeil:
Too rare to be a real issue. I can change it later if I have to.

PR #474, opened 2016-03-20
Adds an AFK timer to the WHO menu for admins. Adds info for normal players.

Cameron653:
Exactly what it says on the title.
It gives admins an AFK timer, letting them see how long someone has been AFK. 

The timer starts ticking upwards after 5 minutes, and as soon as someone moves/does an action, it resets and removes the afk timer from them on the WHO menu.

Gives players info about a CKEY, and tells them if that CKEY is:
In lobby     http://puu.sh/nRgqX/57edac8783.png
Observing http://puu.sh/nRgli/10a1ae89f4.png
Or Playing http://puu.sh/nRgk2/3483494797.png

Based on PolarisSS13/Polaris#1197


Cameron653:
Actually, wait a minute, testing something...

Cameron653:
DNM right now, modifying something to make it a lot more useful for admins.

SpadesNeil:
This seems redundant. We already have an AFK timer when you examine a player. It's also not necessary for normal players to know and potentially exploit this information.

killer653:
It's mainly so people can easily look at the /who menu and see if someone is online, so they could maybe do with them or whatnot. It'd also allow admins to see who has been afk for a long time, check their job, and quickly open the slot back up for active players, if they're in the game and not kicked by the afk timer. I asked in the Skype if people wanted it added in and the responses were positive. There's really not much to abuse, as it doesn't tell you their in game name or anything like that, only that their ckey is afk. It's more of a useful utility than anything else.

antsnap:
If the skype responses were positive, then this should be implemented. I don't think it is that redundant, since we database-ify info only attainable by personal examination. I'm not sure what you mean by exploit, but if people are using the info to steal IDs well, that was already bannable.

Screemonster:
Just like the cryotubes occasionally spitting out ID cards, while it is against the rules for some nob to nip in and swipe it, it's still best not to give them that opportunity.

SpadesNeil:
It opens up an _opportunity_ to steal IDs that I'd rather not exist and creates a gray area in the rules. If someone is idle but not disconnected, it's actually _not_ covered by the rules, and this is intentional, because if someone is idle but not disconnected, it means they are still at the keyboard to _potentially_ react to you roleplaying with them. If they do not respond to your post in a sufficient time frame, you basically have free game to do whatever you were going to do.

For example, if I RP...

/me attempts to snatch John Doe's ID off his suit.
John Doe is online but unresponsive. I give him a full 5 minutes to respond. Nothing.
/me successfully pillages John Doe's ID while he's distracted.

And then I take his ID.

The cryotubes is not the same situation. **That's a bug.**

killer653:
I could change the afk timer to be viewable by admins only if you'd like.

I could also add modify it to let normal players know whos observing and who's playing. Itd be easy enough.

SpadesNeil:
>I could change the afk timer to be viewable by admins only if you'd like.

This is preferable.

>I could also add modify it to let normal players know whos observing and who's playing. Itd be easy enough.

This is also fine.

SpadesNeil:
So "Idle" only displays for admins now, yes?

SpadesNeil:
misclicked

killer653:
Yes.

PR #479, opened 2016-03-25
Adds the tacticool turtleneck to selectable items at round start.

killer653:
Exactly what the title says.

Someone requested it be able to be selected in the preferences menu.

It's got the same function as a regular jumpsuit, and is normally only obtainable from the arcade machine. This makes it so you don't need to keep playing the arcade machine to obtain the jumpsuit.


Screemonster:
It's supposed to be a prize for the arcade machine. Sometimes the RNG sucks if you're looking for a specific thing, but thems the breaks. What would you propose adding to the arcade machine prizes if this were implemented, since this PR would effectively render that prize worthless if people could just take it without any effort?

antsnap:
It's also available if you hack the vending, I think~

Replace the turtleneck with a chameleon jumpsuit and give it a super low drop chance maybe? That might be too much though.

An alternative, just replace the one in the arcade machine with a special color one, like grey instead of black.

killer653:
The tactical turtleneck isn't really the main prize of the vending maching, just one of the possible prizes. 

I'd probably give it the toy katana in exchange if requested, as it's not given out by the arcade machine, but it's listed as a toy.

Also, with the hacking the vending machine, only the toy-vend has it, and I don't think we actually have a toy-vend on station.

killer653:
I'm going to close this, I need this slot open so I can code something else, and this isn't that urgent. I might open it up at a later time, but for the time being, it has no use.

antsnap:
Beep Beep killer that isn't fair. The only file affected would be preferences_gear.dmi. You could just do a merge from upstream and resolve the conflict, which would be a formality since git will auto combine the differences in the dm file.

Issue #478, opened 2016-03-25
Adjust alcohol metabloism to account for player size. Also discuss liver damage.


antsnap:
It doesn't make sense that it takes a micro mouse and a giant dragon the same amount of liquor to get wasted.

I guess this is sort of a balance thing, but it's mostly immersion in my opinion.

Thoughts?

antsnap:
I also want the toxic damage from alcohol to hurt the liver less. I was thinking reducing it by 2/3 but that might be much. Thoughts?

PR #480, opened 2016-03-25
Alcohol Metabolism Changes (Kills liver slower. Size affects drunkenness)

antsnap:
https://github.com/SpadesNeil/VOREstation/issues/478
Howdy,

Alcohol metabolism now depends on your size. From largest to smallest you can tolerate 
(4.5, 2.5, 1, 0.75, 0.5)
Times the drink of a normal person.

I also adjusted the rate at which alcohol can cause tox damage. This should help not-swamp-medbay with the increase in drinking from the larger sizes. Also it's probably more realistic.

### DIFF
--- a/code/modules/organs/organ_internal.dm
+++ b/code/modules/organs/organ_internal.dm
@@ -213,9 +213,13 @@
 				// Damaged liver means some chemicals are very dangerous
 				// The liver is also responsible for clearing out alcohol and toxins.
 				// Ethanol and all drinks are bad.K
+				// BUT for RP and immersion they shouldn't be that bad~
+				// Liver death from alcohol happens over long periods or very acute doses.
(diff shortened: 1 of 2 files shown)


Screemonster:
50 units is like half a bottle. If normal-sized people are chugging down half a bottle of vodka in one go (as they frequently do, then complain about being drunk), damn right they should get hideously, catastrophically wasted.

If people are getting liver damage after a pint of beer, then it makes a lot more sense to nerf the amount of ethanol you get into your system from drinking beer.

Also, with the nerf to alcohol damage, micros are _still_ able to drink more with these changes than they currently are. (nerfed damage to a 3rd, multiplied by 2 for being tiny - that's still only 2/3rds damage)

For really major changes like these it's best to discuss them in issues or a forum thread before powering ahead with a PR.

antsnap:
yeah you are right about the damage. Reverted

antsnap:
I forgot to link thread~

http://forum.vore-station.net/viewtopic.php?f=20&t=582

Issue #394, opened 2016-02-22
Remove Changeling Proboscis Insta-Killing (maybe replace with vore mechanic)


Screemonster:
At present, the only real way for lings in code to work if for them to husk people, which makes them somewhat too murdery for events.

Giving them a vorey means of doing it (such as automatically absorbing genomes from anyone they digest) would fit the theme of the station a lot better.

A way for admins to give characters the changeling ABILITIES without them needing to be an outright changeling would be fantastic. Last time I made an "alien creature" for an event with a nagatail and unathi body, I had to make it a ling and have it eat like half a dozen spawned-in braindeads just to give it the ranged hallucinogen sting and enough chemicals to actually be useful.

OrbisAnima:
it's a relatively simple change to do, I look foward to these events scree!

Screemonster:
Also I just realised that fluffwise a vore-based DNA-stealing stingbeast is basically Scree dialed up to 11 :p

PR #496, opened 2016-04-01
Changelings absorbing people doesn't kill

Cameron653:
Exactly what it says on the title

Does 10 damage, and 60 agony damage. (Enough that you get a cut that heals quickly, and it hurts like all hell. Turns your health doll red, while not actually doing the damage.)


SpadesNeil:
Fixes #394

PR #499, opened 2016-04-02
Allows borgs to digest people. Notified admins when they do so.

Cameron653:
Exactly what it says on the title.
Allows borgs to inject "digestive enzymes" into people which does 3 burn damage per tick.
After they die, if the borgs try to inject them with any chemical again, it gives  them a text box, asking if  they'd like to digest the dead person. If  they do so, it'll tell admins and give the borg 30000 charge (As much as a hyper battery)

Also, ignore the last 5 commits at the bottom. There were a few red and  green lines I was trying to fix from showing up in the files list, as I didn't mean to remove them, and I ended up taking a few  tries. Woo.

Also prevents borgs from killing theirselves when trying to inject someone when they're already at the max chemical limit.

Note: Also makes the sleeper take away less power. It was removing power so fast that scenes were insanely hard to do without constantly recharging.

### DIFF
--- a/code/game/objects/items/weapons/dogborgstuff.dm
+++ b/code/game/objects/items/weapons/dogborgstuff.dm
@@ -191,7 +191,7 @@
 	var/inject_amount = 10
 	var/min_health = -100
 	var/occupied = 0
-	var/list/injection_chems = list("dexalin", "bicaridine", "kelotane","anti_toxin", "alkysine", "imidazoline", "spaceacillin", "paracetamol", "digestive_enzyme") //The borg is able to heal every damage type. As a nerf, they use 750 charge per injection.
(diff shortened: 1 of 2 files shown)


SpadesNeil:
What about the AI laws? Or do we just fluff by saying the K9 units have a defect where the protection law gets interpreted wrongly sometimes?

Cameron653:
For medical borgs, it could be a method to dispose of unwanted bodies instead of putting them in the morgue.
For security cyborgs, their stomach could accidentally "malfunction" while they're  bringing the criminal to the brig, or something similar like that.

Generally, their stomach could malfunction, they might need to digest an assistant to have enough power to inject the captain with life-saving chemicals (Which is where law 3 comes in, which says

"3. Protect: Protect the crew of your assigned space station to the best of your abilities, with priority as according to their rank and role."

There's a few other things that I forget that they could say that would validate them to digest crew. Mainly the protect law loophole.

SpadesNeil:
Kay.

PR #512, opened 2016-04-04
Adds a sharper looking Xenomorph tail

Cameron653:
Tail in action: http://puu.sh/o6eiF/c4bd6b1a8b.png

The current one in game has a backplate, two backspikes, and a tail. This one is just a tail and two backspikes, and looks a lot sharper than the other one, as the other one's tail looks extremely floppy.

Note: Both tails will be able to be chosen, and they both have different names so it doesn't confuse people.


SpadesNeil:
The names need to be more distinguishable from one-another.

Cameron653:
There. I named one
"Xenomorph tail w/ backplate"
and left the other one as
"Xenomorph tail"

PR #522, opened 2016-04-06
Weight gain/loss mechanics

SpadesNeil:
- Ability to gain weight after having eaten too much. Weight gains as nutriment decays (you're digesting and absorbing foodmush).
- Flavortext to accompany the weight of the player.
- Ability to track weight via bodyscanners.
- Ability to lose weight with lipozine, or by starving yourself (running depletes nutriment faster, which means you start burning fat once it's low enough).
- Ability to adjust weight in character setup.
- Ability to adjust rate at which you gain or lose weight in character setup from 100% to 0% (no change).

Future features include exercise equipment and hopefully the ability to keep weight change persistent between rounds.


killer653:
:+1:

PR #523, opened 2016-04-06
Makes robohounds animation faster. Adds in a K69 borg.

Cameron653:
Makes the robohounds animations x2 faster, making them look less floaty when moving around.

Also adds in a "K69 borg", which is adminspawn only (Via VV modification of the borgs icon.) in case they wanted to do an event in which k9 units with a pink color was needed. 

(See: If the BDSM lawset was added.)

K69 sprite made by TalkingCactus(?)


SpadesNeil:
What is the icon name for the BDSM lawset?

Cameron653:
BDSM lawset isn't added. I'm saying that the pink borg sprite could be used "IF" the BDSM lawset is added in. Should've probably clarified. Also, the borg's icon name is K69

SpadesNeil:
I meant the icon name for the BDSM borg but you knew what I meant. K69, ok.

PR #528, opened 2016-04-07
Fix 'replacetext' and add AI boards.

### LINKED ISSUE #515: BDSM Lawset
We talked about it in the admin chat. Yeh we're lewd. That's how it works.

Gonna implement a BDSM lawset so the AI can send in the sexbots.

### LINKED ISSUE #527: "replacetext" BYOND proc
In the next version of BYOND, "replacetext" is a reserved word and implemented as a proc.

Unfortunately bay wrote their own "replacetext" proc since BYOND lacked one. I can't refactor the code to use BYOND's since it's only in beta and the server isn't using the beta. The server can't be updated to use the stable until the code is updated, because "replacetext" is reserved and existing code doesn't compile.

Arokha:
Added BDSM and KOTH AI cores. Admin-spawnable.
Replaces 'replacetext' and 'replacetextEx' procs with 'bay' versions. This is unfortunately messy, file-wise.

Fixes #515
Fixes #527

### DIFF
--- a/code/ZAS/Variable Settings.dm
+++ b/code/ZAS/Variable Settings.dm
@@ -348,7 +348,7 @@ var/global/vs_control/vsc = new
 			var/txt = vars["[V]_RANDOM"]
 			if(findtextEx(txt,"PROB"))
 				txt = text2list(txt,"/")
-				txt[1] = replacetext(txt[1],"PROB","")
+				txt[1] = bayreplacetext(txt[1],"PROB","")
 				var/p = text2num(txt[1])
 				var/r = txt[2]
 				if(prob(p))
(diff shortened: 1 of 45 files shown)


Arokha:
This PR is marked as major due to the fact that it blocks the next BYOND version, and needs confirmation because it does change A LOT of files, though the same files are likely to be changed once the server is updated to clean this back up. More to just make it a seamless update process rather than requiring a code and server update at the same time which just has a higher chance of breaking things.

SpadesNeil:
The proc change isn't just major; it's a blocker. Without this change, we cannot compile any code when Byond updates.

SpadesNeil:
I have a question. Is BYOND's version going to do the same thing as Bay's version? If so, then wouldn't we just want to remove the defined proc that's causing the compile error trying to overwrite BYOND's version?

idk if that makes sense, I just woke up

Screemonster:
If bay's version and byond's version don't use the same arguments it'll be a pain in the ass to convert, renaming the proc will also work right now and be in place ready for the update, rather than waiting for it to break and having to rush it in right at that moment.

SpadesNeil:
All I needed to know.

PR #539, opened 2016-04-09
Console update to "C3" sprites

Arokha:
Bay looks to have been mid-update when we forked in regards to consoles. They had developed some "C3" thing they were switching to. Had all-new sprites for lots of stuff.

This implements those sprites (not the whole 'C3' system). Gives terminal frames department specific looks (for med, sci, eng, and sec anyway). Most of the screens look identical or nearly identical.

The IMPORTANT part, is that this splits the frame, the screen, and the keyboard into different sprites, so if we want to change the look of the terminals in the future, it is easier (HOLOSCREENS!!!), as there are only 5 frame sprites, as opposed to 40whatever all-in-one sprites.

Let me know if there are any weird-looking consoles. I did my best looking around for any mistakes.


SpadesNeil:
Indeed I can already tell this breaks some things including the TV sprite, the guest computer sprite, and probably some other stuff.

PR #541, opened 2016-04-09
Recharge station wording

Arokha:
Let's use like, proper case and stuff.
Also made 'exit' only show up when in a charger.


Arokha:
It's literally a spelling fix. I'll just merge this.

PR #542, opened 2016-04-09
Dogbot QOL Tweaks

Arokha:
Makes dogborgs respect digestable pref. Greys out digestive enzymes when prey are indigestible, and prevents you from digesting indigestible bodies.

Added 'Refuel' button to 'finish' digesting, rather than needing to click an unrelated button to get it to proc the dialog. Only lights up with dead prey who have digestion on.

Fixed Scan button and makes the page refresh when injecting as well.

Can't inject with < 800 energy rather than 750. Still costs 750, but "750 < 750 = false". Could still inject chems and one-click kill yourself. So here's a 50 energy safety net. Keep in mind it's 5 energy per tick on these.

No longer does patient.stat == dead or patient.stat != dead. Don't do direct comparisons on bitfields, and instead, use bitwise operators.

Can't inject innaprov into dead patients (just looked silly). Doesn't auto-inject either, if they are at full health. Feels odd to constantly pump people full of innaprov in endo scenes.

Probably should still be worked into the vore system though.

**Undigestable person in sleeper**
Imgur

**Dead undigestable person in sleeper**
Imgur

**Digestable person in sleeper**
Imgur

**Dead digestable person in sleeper ("Refuel" button)**
Imgur

### DIFF
--- a/code/game/objects/items/weapons/dogborgstuff.dm
+++ b/code/game/objects/items/weapons/dogborgstuff.dm
@@ -191,7 +191,8 @@
 	var/inject_amount = 10
 	var/min_health = -100
 	var/occupied = 0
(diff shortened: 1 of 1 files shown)


killer653:
:+1: Good job!

PR #547, opened 2016-04-09
Beds+Sheets Sprites

Arokha:
Not going to steal these again every 6 months.


Arokha:
items.dmi and objects.dmi have a looot of stuff. just merging this. you can be mad if it breaks everything. I tested it though and... it's... beds.

PR #551, opened 2016-04-10
Increases the max character limit for preferences.

Cameron653:
Increases the max character limit for preferences.
This includes:
**General
Preferences
Limb-Descriptions**

It increases it from 1024 to 3072, as much as a peice of paper per limb/general/pref.

My main concern for this was that it might lag out the client, or cause the client to lag at all.
I tested this out, and it turns out that it doesn't lag at all. Even with the increase of the word limit in prefs.

This allows people that wish to write more words in their flavor text and preferences, meaning that if someone wishes to write a huge blurb for their character, or list a huge amount of prefs in their prefs list, they can.

Note: I marked it major, as it increases the flavor text limit a large amount, allowing people to put down more prerfs and info about their character.


SpadesNeil:
If someone runs out of space now I'm just going to laugh at them.

PR #555, opened 2016-04-10
Adds cable coil to the autolathe.

### LINKED ISSUE #482: Cable coils should be producable on-station.
As it is, there are some in the vendors, a few piles here and there, and the only other way to get more once these have been exhausted is to order in electrical maintenance crates from cargo - if all you're after is some cable and you don't actually need a dozen toolboxes full of screwdrivers this is a bit silly.

I suggest adding them to the list of things to be printed off at the autolathe.

Cameron653:
Alright, this is an extremely... Well, hacky addition, to say the least. #482

So, to start, this adds cable coil to the autolathe at a nearly-free price.

The thing about cable coil is that no matter what I did, it'd always spawn a cable coil stack with 30 pieces in it. Alright, that's fine. Just bump the metal and glass price up by 30, right? Nope.
It gives materials to the autolathe _per piece_ of cable coil. Not per stack. This means that no matter how small you put the price of the cable coil, you'll always get 30 times the amount of materials if you place it back in.

Now, this is a _huge_ problem, as if it's 10u of metal, 10u of glass, that becomes 300u of each, which can be used to create 30 coils of each, which can then be redeemed for 3000u of each, etc etc...

To avoid this, I made the price of a cable coil 1 metal, 1 glass. This means that using it will have such a small effect on the total metal/glass count it'll be a waste of time to do so.

Like I said, it's extremely hacky, but it works. Without this, there's really no way that I could see to get the autolathe to create cable coil without it bugging out, as seen here: 
http://puu.sh/ocLBf/9100424029.png

Resolves #482


Screemonster:
The hacky part doesn't sound good. Rods and sheets don't have this problem.

should the cable not have the is_stack = 1 part that the other stacks of materials have?

Cameron653:
>should the cable not have the is_stack = 1 part that the other stacks of materials have?
I tried that, but it comes out like this http://puu.sh/ocLBf/9100424029.png
I really couldn't find any other way to easily get it to make stacks of cable without completely messing up.

SpadesNeil:
I don't want a hacky solution to this. Cable coils can be ordered through cargo for the moment and I'm not going to add this in its current state.

SpadesNeil:
If it can be done with metal sheets it can be done with cables.

Leshana:
Hi, your autolathe recipe needs to have is_stack = 1, that takes care of the problem.
However, while researching this I noticed that it is possible to use the autolathe to construct stacks bigger than the normal stack size maximum.  Is that a problem, or a cool feature (tm) ?

killer653:
Iirc, I tried using the is_stack and it caused bigger stacks with no limit. This caused errors and the game to send an admin message saying someone was trying duplicate items. 

Bigger stack: http://puu.sh/ocLBf/9100424029.png

Leshana:
@killer653 Yes that is mostly true.   It allowed bigger stacks up to the amount of resources available.  There is a hard-coded anti-cheating check which catches stacks > 100x
To avoid this annoyance I added code to limit the allowed number of items to the normal maximum stack size of the type.  See #637

PR #557, opened 2016-04-10
Job Whitelist

### LINKED ISSUE #493: Jobs whitelist
We need a proper jobs whitelist which behaves like the species whitelist in a config file.

New whitelisted jobs will include things like...

- CentCom staff
- Syndicate listening outpost crew
- Mimes
- Clowns
- Blue Shield (if it's ever re-added)

Until this whitelist is added properly, I cannot add these whitelisted positions.

Arokha:
This allows any job to be set to require whitelisting by adding the var
"job_whitelisted = 1" to the settings for the job. It also adds a "Unique"
department, and an example job, "CentCom Observer", in a new special.dm
file for the special whitelist/admin jobs. There's no config file for
WHICH jobs are whitelisted yet.

Admins always ignore the whitelist.

This makes the job priorities screen ugly, sorry. It will look nicer when
that third column has more jobs in it, or someone who knows more
HTML-as-tortured-by-Byond can take a crack at it.

**_THINGS YOU SHOULD KNOW:_**
You must create jobwhitelist.txt in the config folder. After updating,
there will be a copy in config/example.

There is a new config file option, USEJOBWHITELIST, that you need to add
to enable job whitelists. Until then, this does not do anything. Do not
use USEWHITELIST, as that is the old version that only applies to head
roles.

You can whitelist any job by going to the appropriate dm (medical.dm,
captain.dm, etc) and adding "job_whitelisted = 1".

I am 'fairly' sure this is robust. Someone extremely sneaky might figure
out a way around it, but I can't think of a way. The randomizer skips you
properly I believe, and you can't pick whitelist jobs when latejoining.

Comments apparently are not ignored in config files so if you do something
like #arokha - All, it will still give all jobs whitelisted.

Resolves #493

### DIFF
--- a/code/game/jobs/whitelist.dm
+++ b/code/game/jobs/whitelist.dm
@@ -1,5 +1,6 @@
 #define WHITELISTFILE "data/whitelist.txt"
 
+// Original 'job whitelist', old style (for exec jobs).
 var/list/whitelist = list()
 
 /hook/startup/proc/loadWhitelist()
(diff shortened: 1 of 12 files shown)


Arokha:
Reasons for file changes:
VOREStation.dme - Adds special.dm job file
code/controllers/configuration.dm - Adds config file option to enable job whitelist
code/game/jobs/job/job.dm - Adds job_whitelisted variable
code/game/jobs/job/special.dm - Adds example CC observer job, whitelisted
code/game/jobs/job_controller.dm - Adds checks for whitelisted jobs and avoids giving them to people who aren't whitelisted
code/game/jobs/jobs.dm - Adds 'Unique' department and CC observer job
code/game/jobs/whitelist.dm - Adds proc to check if someone is on the whitelist, read the whitelist config file, and improved species whitelist for fun
code/modules/client/preferences.dm - Prevents picking job priorities for jobs you can't take
code/modules/mob/dead/observer/observer.dm - Fixed a random incorrect message. It's not 30 minutes, and our rules are different.
code/modules/mob/new_player/new_player.dm - Controls the latejoin job list, hides whitelisted jobs you don't have
config/exmaple/config.txt - Gives example of the USEJOBWHITELIST option
config/example/jobwhitelist.txt - Example job whitelist file

SpadesNeil:
This is awesome. Now I can re-enable jobs like mime or clown.

Also before anyone asks
# I am not adding any standard station jobs to this and I never will because that is a terrible idea that chips away at server population. That means I am not adding Captain or HoS to this. If any of you complain, I'm gonna scream.

That being said I can also add in some new and unique jobs that simply weren't available prior.

code/game/jobs/whitelist.dm
```diff
@@ -30,12 +62,16 @@ var/list/whitelist = list()
 	else
 		alien_whitelist = text2list(text, "\n")
 
-//todo: admin aliens
 /proc/is_alien_whitelisted(mob/M, var/species)
 	if(!config.usealienwhitelist)
 		return 1
-	if(species != "vox" && species != "Vox" && species != "diona" && species != "Diona" && species != "skeleton" && species != "Skeleton")
```
SpadesNeil: This looks like the problem.

ForFoxSake: Already testing a fix for this now.

Arokha: Already fixed it

PR #558, opened 2016-04-10
Fix for #494

### LINKED ISSUE #494: Exosuits don't register ID's placed in wallet
As the title says - when you have an ID in your wallet, it provides access for you, just as placing it in a PDA... but exosuits don't register them.

Arokha:
Lets you use wallets on exosuits (and rigsuits because I was not 100% sure
what the person reporting it was referring to, and both needed fixing).

Fixes #494

### DIFF
--- a/code/game/mecha/mecha.dm
+++ b/code/game/mecha/mecha.dm
@@ -702,15 +702,18 @@
 			else
 				user << "You were unable to attach [W] to [src]"
 		return
-	if(istype(W, /obj/item/weapon/card/id)||istype(W, /obj/item/device/pda))
+	if(istype(W, /obj/item/weapon/card/id)||istype(W, /obj/item/device/pda)||istype(W, /obj/item/weapon/storage/wallet))
 		if(add_req_access || maint_access)
 			if(internals_access_allowed(usr))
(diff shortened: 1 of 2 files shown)


Arokha:
Ignore, this wasn't what he meant, but still should have been fixed. Will include in full fix.

Arokha:
Nevermind I was right and already fixed in the first PR.

SpadesNeil:
So this fixes preventing players from getting into exosuits while their ID is in a wallet?

Arokha:
Yes, and allows you to whap exosuits to get the maintenance menu with them in a wallet, and also open the maintenance panel on rigsuits with ID in a wallet. You can pretend it's the wallet upgrade.

SpadesNeil:
I don't think that is what they were asking to be fixed.

Arokha:
What do you think it was?

SpadesNeil:
>Don't actually remember if I tested that. I meant for providing access to enter the exosuit.

SpadesNeil:
Oh, nevermind.
> Having a secured mech (security mechs etc) and getting in while your ID is in a wallet

PR #578, opened 2016-04-13
Move/rename gateway map load text file

Arokha:
maps/RandomZlevels/fileList.txt >> config/awaymissions.txt

Included example in examples.


Arokha:
Bug because it sure does bug me hyuck hyuck.

SpadesNeil:
Leaving this open juuuuuuuuuuuust until the next server update so I don't forget to move the file.

Issue #136, opened 2015-10-14
Forks icons are wrong


SpadesNeil:
Sprites are wrong directions.

Leshana:
What do you mean sprites are wrong?

SpadesNeil:
They face the wrong directions.

SpadesNeil:
Keeping this issue open because the directional icons are still wrong.

nightwingthedragon:
Eating soup with a fork tested and confirmed in issue #196

PR #592, opened 2016-04-14
Fixes #99, #136, #159

### LINKED ISSUE #99: Wiggle cooldown/randomness is much too sparse.

### LINKED ISSUE #159: IPCs and genetic damage from slimes
I don't really know what to say about this or if it's something that can be repaired.
Prior to update, IPCs took general damage from slimes. Now they seem to take invisible genetic damage (no genetics to damage) that they can't count or feel and nobody else can see or repair and eventually if they are unaware of it they will simply suddenly die. Everyone else will see a perfectly healthy looking but dead IPC. There's no surgery or anything to be done on an IPC corpse so it just.. is dead.
Had this happen when I was playing my IPC in sec and helping handle a massive slime outbreak, assumed I would just get damage welded off if it got too bad but just sort of suddenly died to the confusion of many. 
(At the time it was agreed that CentComm was to send a replacement model in and I was to wait a few minutes after respawn timer and then spawn back in, as a 'solution', though my dead self was on the manifest the rest of the round.)

Arokha:
Resolves #99 - Wiggle cooldown/randomness is much too sparse
- 2.5 second delay, no prob() call for random chance

Resovles #136 - Forks icons are wrong
- Fixed for forks, just copypasted the correct sprite

Resolves #159 - IPCs and genetic damage from slimes
- Does brute damage, which is what they do to anything that is not /carbon/human, like simple mobs. Was consistent. Also applies to borgs.


code/modules/mob/living/carbon/metroid/powers.dm
```diff
@@ -42,13 +42,13 @@
...
 			UpdateFeed(M)
 
-			if(istype(M, /mob/living/carbon))
+			if(istype(M, /mob/living/carbon) && !is_mechanical(M))
 				Victim.adjustCloneLoss(rand(5,6))
 				Victim.adjustToxLoss(rand(1,2))
 				if(Victim.health <= 0)
 					Victim.adjustToxLoss(rand(2,4))
 
-			else if(istype(M, /mob/living/simple_animal))
```
Arokha: Checks here if they are mechanical. This returns true for cyborgs, and the "machine" species, as well as AI characters.

code/modules/vore/belly/belly_stomach.dm
```diff
@@ -131,50 +131,54 @@
 
 // @Override
 /datum/belly/stomach/relay_struggle(var/mob/user, var/direction)
-	if (!(user in internal_contents))
```
Arokha: Drops out if recent_struggle is true, which is set true a few lines lower.

code/modules/mob/living/carbon/carbon.dm
```diff
@@ -66,11 +66,14 @@
 
 //Vore code, struggle stuff
 /mob/living/carbon/relaymove(var/mob/user, var/direction)
+
+/* This stuff is on bellies. 'recent_struggle' is not defined on carbons.
```
Arokha: Commented out because it's not in the right place.

code/modules/mob/living/carbon/metroid/powers.dm
```diff
@@ -42,13 +42,13 @@
...
+			else if(istype(M, /mob/living/simple_animal) || is_mechanical(M))
```
Arokha: Checks here if they are mechanical. This returns true for cyborgs, and the "machine" species, as well as AI characters.

code/modules/vore/belly/belly_stomach.dm
```diff
@@ -131,50 +131,54 @@
...
-		return  // User is not in this belly!
+	if (!(user in internal_contents) || recent_struggle)
```
Arokha: Drops out if recent_struggle is true, which is set true a few lines lower.

PR #603, opened 2016-04-18
Adds scoutbot sprite to robots.dmi

### LINKED ISSUE #599: Robotic recon drone with mounted camera
I'd like to have a small drone with 1hp controlled by a positron brain, similar to spider bots in the old code, that serve no other purpose than to go through a gateway or into other hazards and send video feedback to the station.

They are equipped with nothing but an on board portable camera that can be watched from a monitor much in the way headset and helmet cams work. Maybe a multitool to calibrate the gateway so they can come home. Literally nothing else.

Screemonster:
Sprite addition for whatever we cook up to resolve #599 - it's a little four-wheeled curiosity-looking thing.
Icon state is "scoutbot".


SpadesNeil:
This is just the icon so the issue is not resolved.

PR #605, opened 2016-04-18
Gateway sound

### LINKED ISSUE #600: Give the gateway a sound effect on use
Makes the mechanic more juicy.

Arokha:
VWIIIZSHHHH.

Resolves #600

### DIFF
--- a/code/modules/awaymissions/gateway.dm
+++ b/code/modules/awaymissions/gateway.dm
@@ -136,6 +136,8 @@ obj/machinery/gateway/centerstation/process()
 			M.loc = dest.loc
 			M.set_dir(SOUTH)
 			use_power(5000)
+			M << 'sound/effects/phasein.ogg'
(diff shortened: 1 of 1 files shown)


SpadesNeil:
centeraway/process() needs this too!

Arokha:
Um but process() is called each time the process controller iterates over it. That’d play every controller tick. c.c; You sure about dis? Unless process() here means something entirely different. Lemme look.

Arokha:
Yeah if I add it to process() it’ll play every tick. Dis seems like bad ideas.

SpadesNeil:
Sorry I'm an idiot. I mean

 /obj/machinery/gateway/centeraway/Bumped(atom/movable/M as mob|obj)

You only did:
 /obj/machinery/gateway/centerstation/Bumped(atom/movable/M as mob|obj)

Arokha:
Fwoop. Did.

PR #612, opened 2016-04-19
Borg Whispering

Arokha:
Adds whispering to borgs. Was annoying to RP without that. I'll have to add subtle emotes too but that's a lot more complicated. Also made it turn off your light when resting. Also think I caught a bug about the dogborg resting anim.


SpadesNeil:
Whispering not working is a severe bug since it really gets in the way. Adjusted tags appropriately.

Lights turning off is minor feature.

Issue #487, opened 2016-03-28
New vore mob icons


SpadesNeil:
We need some vore mob icons added in.

There should be an icon for full and empty, as per specifications of #486 

Note that vore mobs can be up to 64x64 pixels, but must be placed in a separate file.

All mobs require sprites facing north, south, east, and west while alive.
All mobs require an empty sprite set, at least one full sprite set, and a dead sprite.

All vore mob icon files should go under:
**VOREstation\icons\mob\vore.dmi** and **VOREstation\icons\mob\vore64x64.dmi** for icons larger than 32x32 pixels.

Some things I'd like to see us start with...
- Snakes
- Xenomorphs
- Deathclaws (yus yus... shut up)

Screemonster:
64x64 of the alpha male and mother deathclaws from quarry junction? Yus.

SpadesNeil:
Vore mobs are coming. Need moar icons. I also need a giant 64x64 albino space carp. Don't ask why.

PR #615, opened 2016-04-20
Adds vore64x64.dmi with large snake mob icon

Screemonster:
As per commit and description. Adds icons for a large snake mob. Has directional sprites for full and empty states, and a dead sprite.


Screemonster:
Is for #487 but this PR won't resolve the issue on its own.

PR #643, opened 2016-04-30
Defibrilator and Crash Kit

Arokha:
Adds an automated defib vest and a crash kit for medical (2 inaprov autoinjectors, 1 vest).

Proper use of defib vest:
1) Inject person with inaprov (helps stop them from re-dying, optional)
2) Activate vest (click in hand)
3) Remove their uniform, and put the vest on them
4) Wait, hope

The vest has enough for 3 attempts at re-starting their heart. You can upgrade the battery by replacing it with a screwdriver. After 3, it's much less likely to work. After 6 minutes, patients cannot be revived.

In theory, the most you can revive someone from is -155 HP, if they have exactly 35 oxy and 10 burn and 10 brute over the -100 limit.


Arokha:
Ace said to so I'mma merge it

PR #645, opened 2016-04-30
Adds TF to the new vore system.

Cameron653:
Same as the last TF setup, but made to work with the new-vore system. No differences other than that.


Arokha:
Made slight changes for speed, and to hide TF options for non-human preds.

PR #662, opened 2016-05-02
Medical Vore Upgrade

Arokha:
- Added two drugs, ickypak and unsorbitol
- Added recipies for them (documented on wiki)
- Added prey scanning to health scanner and body scanner
- Made bellies show up in lowercase in more places as appropriate
- Added ickypak and unsorbitol bottles, for admin spawn (or map use?)
- Moved message for vore ejecting to belly rather than anywhere that calls it

I'll also pretend the purple crash kit medkit I added last update is part of this, yis.


Arokha:
I'm going to merge this, somewhat experimentally, because I'm a bit concerned about those other two commits in there and I'm unsure if this is going to destroy the universe. -.o And I kinda need to know when it's a time I can fix it if everything breaks and I'm not at work etc.

PR #669, opened 2016-05-02
Fix Runtimes

Arokha:
No idea why these are happening but this cleans it up. The silly way they were doing it was dumb anyway.


Arokha:
A fix for my previous fix, just merging.

Issue #629, opened 2016-04-27
Using a reset board on a dogborg makes them vanish, and charging with >100% resets to 100%.


Screemonster:
Request for Help:: GuiltyBeans/(Cobalt)(?) (PP) (VV) (SM) (JMP) (CA): Also, Scree, (?) I found a bug. I don't know how to report on the Github. Found two actually. One is, when trying to use a reset module board on a dogborg, it makes their sprite vanish. I can help you recreate it to see. Also! Second one, I ate arbon, gained 200% charge on my cell. Stepped into a recharge station and it knocked me all the way back to 100% 
  to GuiltyBeans/(Cobalt)(?): I'll make issues for them.
  [Player PM] GuiltyBeans/(Cobalt)(?): Yay! Thank you! They told me to report them but I don't know how ;w;
  to GuiltyBeans/(Cobalt)(?): Steps to recreate the sprite bug? Just use a reset board on them like normal?
  [Player PM] GuiltyBeans/(Cobalt)(?): Yeah! When you have them opened, Lasshseeki was doing an upgrade on me. When he used the board after putting in a new battery, I vanished.

Cameron653:
The way the robohound sprites work is that when the person selects their
module, they move 16 pixels to the left so their sprite is in line with the
tile. This could probably be fixed by making the reset board look and bump
their sprite 16 pixels to the right if they get reset as a robohound.

Arokha:
Rather than a specific amount, I’d suggest just undoing the pixel.x and pixel.y values (using ‘initial’).

Cameron653:
Turns out, it left the icon at "icons/mob/widerobot.dmi". Should be an easy enough fix.

Cameron653:
Fixed the first half of this issue. Didn't really bother all that much with the second half, which is the borgs going to max charge when they walk into a charging unit, because there's really no reason for them to go into one in the first place, and it's so minor that it's not a big problem.

SpadesNeil:
Please make a separate issue for the charge problem.

Really, borgs should not be able to go over 100% charge anyway. That is an exploit and a bug.

Cameron653:
The borgs being able to go above 100% charge was originally put in to help for long scenes, since borgs will lose power 24/7, even if they have no modules selected, which means you would have to run back and forth from the charger during long scenes

PR #671, opened 2016-05-02
Fixes cyborg sprite when reset module is used #629

Cameron653:
#629 
Exactly what it says on the tin. 

I was also going to fix the "Resetting the borg to normal charge" part attached to the issue, but I didn't see exactly where it could be fixed without causing issues. (Ex: Borg infinitely charging with no stop), and this is the biggest problem in that issue request. Borgs walking into a charger with above max charge really isn't a huge issue.


Arokha:
I'd generally agree with your sentiment on that 'max charge' issue, yeah. It's just a helpful feature. The game isn't getting designed around it.

PR #678, opened 2016-05-03
Fix for secure crates.

### LINKED ISSUE #461: Abandoned crates should explode when shot

Screemonster:
Removed an unnecessary call to the parent function in secure crates that was causing it to just spill the contents like any other generic locker instead of actually doing the tamper_proof checks.

Fixes #461.


Screemonster:
... the alternative solution would be to just pile the tamper_proof check and var into _all_ closets and children thereof. It'd mean running an extra if statement when blowing open crates/lockers that never have the tamperproof property by default, but at the moment _only_ things under /closet/crate/secure can ever be tamperproof effectively. Not that this is ever going to be an issue unless an admin wants to be cheeky and varedit tamper_proof to 1 on things that don't normally have that property.

PR #682, opened 2016-05-03
Fix for scan->cryo->rejoin->scan

Arokha:
Makes cryo search people for health implants, and deletes the associated
record from the scanning database whenever the person is deleted for cryo
purposes. Also made the proc for finding preserved items less expensive.

Surely there was an issue about this at some point? Maybe not anymore?

### DIFF
--- a/code/game/dna/dna_modifier.dm
+++ b/code/game/dna/dna_modifier.dm
@@ -13,7 +13,7 @@
 
 	// Stuff for cloners
 	var/id=null
-	var/implant=null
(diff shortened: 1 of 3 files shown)


SilverTalismen:
There is, As it is if you get a scan and switch characters, if someone tries to clone your prior character it rips you out of the current one

Arokha:
Well I mean I know it is 'an issue' but I meant a github issue. X3

SpadesNeil:
So this deletes clone records when you cryo? Should probably change the title to say that.

PR #684, opened 2016-05-04
Insole for Micros

### LINKED ISSUE #280: Stuffing micros inside shoes
I would LOVE the ability to put micros inside footwear, maybe by adding a slot to the footwear were only micros are allowed or being able to add a micro to a shoe like it was an accessory like a medal to a jumpsuit, they should not be able to be spotted by just examining the person like you would to see a medal.

Arokha:
### Alright Pawfiends!

Now you can put micros in footwear (and this adds the general framework to add them to other stuff as well, but just works on shoes for now).

1. Grab a micro.
2. Click footwear.

**Click footwear in hand to remove micros.**

Micros can squirm in footwear and occasionally get emotes when you're walking. They can also resist out. If you're just holding the shoes or they're on the ground it's fairly easy, but if you put them on, or are already wearing them and take a step, it will pin them back down!

Also added resisting out of dead mobs, for good measure. Was unrelated but, you know.

Resolves #280


SpadesNeil:
@Rixunie

PR #690, opened 2016-05-04
The return of bellytext in examines

Arokha:
This wasn't intentionally removed, but sorta half-broken. So now you can edit your belly's examine text messages, and by that I mean the messages it gives to people who examine you when you have someone/something in a belly.

"Their taurguts are swinging a bit low."

etc.


Arokha:
Need to update, and this is just the return of an old feature that was accidentally removed.

PR #696, opened 2016-05-05
Using New() instead of initialize() for closets and filing cabinets

SpadesNeil:
I don't know a lot about this topic, but I know that on away missions, New() works and initialize() does not work for spawning objects in maps generated at runtime.


Arokha:
Initialize is something called, I think, when a world loads on all the objects in the world. New is called whenever a new that-thing is created. I'm unsure of the difference in regards to your away mission.

SpadesNeil:
I just know that the way initialize works, it doesn't function correctly for away missions. It doesn't get called IIRC.

